PR #10, opened 2019-03-25
Model & Seed work

thomasv314:
- Adds initial migrations. Run `python manage.py migrate` to setup state in DB.
- Adds a seed command. Running `python manage.py seed` will create a default admin user and 10 random participants.
- Default admin user can be logged in with username `admin`, and password `password123`.  After logging in as admin creating/editing models is available by visiting http://localhost:8000/admin/


bkarlovitz:
Looks great.

PR #22, opened 2019-04-09
Add participant search component

marcelkooi:
Still need to figure out how we want to dynamically render components/pages but this is the first start on the Participant Search component

### DIFF
--- a/frontend/.stylelintrc
+++ b/frontend/.stylelintrc
@@ -1,4 +1,3 @@
 {
-  "extends": "stylelint-config-recommended",
-  "extends": "stylelint-config-recommended-scss"
+  "extends": "stylelint-config-recommended", "stylelint-config-recommended-scss"
 }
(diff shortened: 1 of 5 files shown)


marcelkooi:
@grahamalama Do you have an opinion on how we should dynamically render different pages/components? Do we want to pull react-router into this project?

grahamalama:
@marcelkooi react-router totally makes sense. I'll start a branch locally to at least get a `/login` route and `/route` for main content.

And yeah, some state management does make sense. Depending on how complex we envision this app becoming though, we might be able to get away with Mobx? Thoughts?

marcelkooi:
> react-router totally makes sense. I'll start a branch locally to at least get a /login route and / route for main content.

Sounds good!

> Depending on how complex we envision this app becoming though, we might be able to get away with Mobx? Thoughts?

I haven't worked with Mobx but I'm happy to try it out 👍

PR #27, opened 2019-04-14
Employee api

MikeyManoguerra:
employee and employee role


bkarlovitz:
employee api endpoint working via django rest framework, so merging to trunk

PR #55, opened 2019-05-01
Navbar enhancements

livetodeliver:
Worked on alignment of navbar and style of Welcome heading.

### DIFF
--- a/frontend/src/scss/navbar.scss
+++ b/frontend/src/scss/navbar.scss
@@ -3,4 +3,10 @@
         justify-content: space-around;
         flex-direction: row;
     }
+    &__gs{
+        background: white;
+        color: #076375;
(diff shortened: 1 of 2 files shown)


frontend/src/scss/navbar.scss
```diff
@@ -1,8 +1,12 @@
 .navbar{
     &__toolbar{
-        justify-content: flex-end;
+        justify-content: space-around;
```
grahamalama: I think `justify-content: space-between;` might work a little bit better for this. What do you think?

livetodeliver: I think I had been going back-and-forth between the two when I originally made the changes, and can't quite remember why I chose space-around instead. Now that I look at it again, I would agree that space-between works better.

PR #76, opened 2019-05-24
Nested services

MikeyManoguerra:
this branch uses the programs and services models to return a list of programs with the services populated, and a list of services with the program names populated, respectively


core/tests/programs.py
```diff
@@ -0,0 +1,51 @@
...
+from core.programs.models import Program
+from rest_framework import status
+import random
+import json
+
+class ProgramsTests(BaseTestCase):
+    fixtures = ['services.yaml', 'programs.yaml']
+    def setUp(self):
+        super().setUp()
+        call_command('loaddata', 'core/fixtures/services.yaml', verbosity=0)
```
thomasv314: are these `call_commands` needed if we define the `fixtures=[]` attribute?

bkarlovitz:
Looks good after the test cleanup 👍

core/tests/programs.py
MikeyManoguerra: nope, fixed!

PR #94, opened 2019-06-18
Updates for service event data

thomasv314:


core/migrations/0001_initial.py
```diff
@@ -1,7 +1,8 @@
-# Generated by Django 2.2 on 2019-04-20 21:41
```
mattusifer: are these migrations just consolidated because we haven't released yet?

mariekers:
Should these be changed to visit events?

PR #97, opened 2019-06-22
Service status

MikeyManoguerra:
updates ServiceEvent view to include service_availability. for now it just includes the whole service object for each, but can pair down if necessary with seperate serializer. i did not get the docker app running unfortunately, this commit does not reflect it but i did add the port to the db object in .yml to no effect. I ll read up on the docs and see if i'm making a dumb mistake, yall directions were thorough.


MikeyManoguerra:
What this PR does: 
-visits endpoint now returns service availability status data along with.
-service_event tests commented out, pending removal

What this PR doesn't do:
-Upate service_event_data to become visit_data
- make any migrations away from our current tables into tables that match prevention point forms

tantanchen:
so was this PR rejected? Is there anything outstanding that needs to be done here?

PR #98, opened 2019-06-22
Bump material to v4

grahamalama:
Bump material version to 4.x. This will aid in development as we'll be able to more-quickly build off examples in Material's docs.


robb33:
this version works pretty smoothly so far.....

PR #106, opened 2019-07-07
potential service_availability model, might need to revisit naming of…

Nkdiaz:
… service vs. program availability. If it needs to be changed it is a simple naming fix. Will talk on Tuesday.


mariekers:
Mind changing it to program availability?

PR #112, opened 2019-07-17
Fix README

mattusifer:
Remove outdated /app/ directory from readme.

I tested this and confirmed that these are the correct steps.


mattusifer:
This is a very small change, so I'm going to merge it for now. If people encounter any more trouble with these steps we can modify them in the future.

PR #115, opened 2019-07-26
Program availability endpoint

MikeyManoguerra:
can get list of availabilities and also update them. Even though Natalya and i discussed using dateTimeField() instead of timeField(), I kept the latter in bc I think having dates that reflect an update date instead of the day it is used for would be confusing?  because these values are only semi regularly updated. if i am wrong i can change it back, date handling is confusing.


MikeyManoguerra:
> Looks like you'll need to rebase and regenerate migrations before merge

I am going to actually remake the PR on the dev branch, i think thats the practice now

PR #119, opened 2019-08-06
Search participant

robb33:
Update to the frontend


robb33:
a lot of these changes are FE...I can't see anything that would be a show stopper...would like to push this up so we can move forward

mariekers:
@bkarlovitz or @Nkdiaz - would either of you mind reviewing this?

robb33:
the sooner the better....I really want to set a baseline here...

robb33:
we need to stop this request....James established a cleaner baseline....I can move some of this over without conflict

PR #120, opened 2019-08-08
Fake data

hhummel:
Create fake data for models exposed in the REST API


hhummel:
Looks like there's some unit testing to fix!

mariekers:
I think @mattusifer and @MikeyManoguerra might be better at checking this

hhummel:
One improvement that occurred to me is to add a random.seed() so the seeded values are reproducible.

hhummel:
I admit I had a little fun with it.

PR #122, opened 2019-08-18
Make fake visits and front desk events all today, leave some FDE's st…

hhummel:
…ill pending, (ARRIVED, CAME_BACK state), increase default # of visits

I changed the fake-date seed.py so the front desk events would be visible in the queues. There are only 3 real changes:  Make the Visits and FDE's created_at using now(), change the random network that makes the FDE's so that some visits are in a pending state, (ie ARRIVED or CAME_BACK), and increase the number of visits so each Program has something in the queue

### DIFF
--- a/core/management/commands/seed.py
+++ b/core/management/commands/seed.py
@@ -1,5 +1,5 @@
-import random, pytz
-
+import random
+from django.utils import timezone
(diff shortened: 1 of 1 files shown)


swearingbird:
Looks good!

PR #125, opened 2019-08-23
Timezone

MikeyManoguerra:
I hope this is what was needed. i manually tested it and the timestamps reflected a 4 hour diff between this update and UTC. not sure the best way to write an actual test though

### DIFF
--- a/core/management/commands/seed.py
+++ b/core/management/commands/seed.py
@@ -1,10 +1,5 @@
-<<<<<<< HEAD
-
-
 import random, pytz
-=======
 import random
->>>>>>> 6f221598814747d70887782a30a01a74620f6651
 from django.utils import timezone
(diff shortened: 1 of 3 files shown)


core/management/commands/seed.py
```diff
@@ -204,10 +199,7 @@ def create_visits(output=True):
     for _ in range(DEFAULT_NUMBER_VISITS):
         v = Visit()
         v.participant = random.choice(participants)
-<<<<<<< HEAD
         #v.created_at = pytz.utc.localize(fake.date_time())
-=======
->>>>>>> 6f221598814747d70887782a30a01a74620f6651
```
mattusifer: Artifact from a merge conflict?

MikeyManoguerra: looks like it

PR #140, opened 2019-09-23
Automated testing 128

jkeary:
This pr shows how to write unit and integrations tests on the front end.  Its establishing a baseline for how to create tests on components


hhummel:
Looks good!

Issue #143, opened 2019-09-24
Empty queue error message


mariekers:
Update empty queue error message - stop displaying 404 error

Done

- [ ] empty queue message: "queue is empty"

mattusifer:
@mariekers I've opened a PR (#151) to return an empty array instead of a 404 for an empty queue. The front end will need to be updated to display the "queue is empty" message if the response is empty.

mariekers:
Added FE issue to resolve - thanks for the heads up, Matt!

PR #151, opened 2019-10-06
Return an empty array for an empty queue

mattusifer:
Resolves https://github.com/CodeForPhilly/prevention-point/issues/143

This PR also fixes the indentation in queue/views.py


MikeyManoguerra:
Looks good! I just merged something ahead of this, so you you may have to include more fixtures in your test.

PR #158, opened 2019-10-24
Homepage waitlist 86

hhummel:
Connect queueStore to apis, and use the storeContext to populate data in AllQueues.  Unit and integration tests for AllQueues.


frontend/src/components/AllQueues.js
```diff
@@ -80,35 +88,86 @@ function QueueTab(props) {
...
+            className={classes.queueTab}
+            queueData={queueStore.queueIds["BIENSTAR"]}
+          />
+          <QueueTab
+            className={classes.queueTab}
+            queueData={queueStore.queueIds["SKWC"]}
+          />
+        </Tabs>
+      </AppBar>
+      {value === 0 && <QueueTable queueData={queueStore.queueIds["TESTING"]} />}
```
marcelkooi: Could we map each `value` with a `queueId` instead?

```js
const valueToQueueId = { 0: 'TESTING', 1: 'CM', 2: 'SSHP', ... , 8: 'SKWC' }

...

<QueueTable queueData={queueStore.queueIds[valueToQueueId[value]]} />
```

And perhaps we could rename `value` to `tabValue` to be more descriptive/explicit

frontend/src/components/QueueTable.js
```diff
@@ -98,20 +106,17 @@ function QueueTable({ queueData: { name, rows } }) {
             render: ({ id }) => <NotesButton id={id} />,
           },
         ]}
-      />
-      <NotesDialog
-        visibleDialog={visibleDialog}
-        toggleVisibleDialog={toggleVisibleDialog}
```
marcelkooi: Is `NotesDialog` supposed to be added back in somewhere?

frontend/src/routes/ServicesQueues.js
```diff
@@ -1,87 +1,141 @@
...
+  hide: {
+    display: "none",
   },
   drawer: {
     width: drawerWidth,
+    flexShrink: 0,
+  },
+  drawerPaper: {
+    width: drawerWidth,
+    background: "#F2FCFF",
```
marcelkooi: Can we add this color to the theme and read it from there instead?

marcelkooi:
Looking good so far! Just had a few requests/questions.

core/queue/views.py
```diff
@@ -30,8 +30,11 @@ def retrieve(self, request, program_id=None):
             created_at__date=datetime.date.today(),
         )
 
+        #todays_visit_data = VisitWithPopulationSerializer(
+        #    visits_queryset, many=True
+        #).data
```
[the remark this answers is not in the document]
hhummel: Yes, I'll take that out.  I added the context to fix an error I received.  In the second case, I would have thought a context would be needed as well, but no error appears.

core/queue/views.py
```diff
@@ -49,6 +52,9 @@ def retrieve(self, request, program_id=None):
                 visit_event_data = FrontDeskEventForQueueSerializer(
                     front_desk_events_queryset, many=True
                 ).data
+                #visit_event_data = FrontDeskEventForQueueSerializer(
+                #    front_desk_events_queryset, many=True, context={'request': request}
+                #).data
```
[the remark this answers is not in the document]
hhummel: Rather than cut lines 55-57, I cut 52-54. As above, there should be a context, although it didn't throw an error. In my opinion it should and I would like it to be consistent with the case above

frontend/src/components/AllQueues.js
hhummel: I renamed "value" "tabValue".  

Regarding reversing the map, the mapping you describe is already in queueStore in queueStats.  We could simplify considerably by setting the tab numbers in store and just setting them up in the component by number. Making the tabs is just a for loop, and we don't need all the if's at the end. 

I did it like this because I thought it would be better to set the order in the component.

frontend/src/components/QueueTable.js
hhummel: I'm not sure how it got cut.  You made it, didn't you? Is this where it should be?

frontend/src/routes/ServicesQueues.js
hhummel: I don't know much about how to do that. Can you help?

frontend/src/components/AllQueues.js
hhummel: I didn't much like that queueIds and queueStats have overlapping data, but I thought it made things easier to understand

frontend/src/components/QueueTable.js
hhummel: Putting it back in

hhummel: Fixed. Sorry for cutting it

frontend/src/components/AllQueues.js
hhummel: Open to changing it

hhummel: It's a lot better simplified. Changed it.  Getting rid of queueIds in queueStore, i think

frontend/src/routes/ServicesQueues.js
hhummel: I put the background color into the styling on App.js and commented it out in ServicesQueues.js.  Could you take a look and tell me if I did it correctly?  Winging it on this.

frontend/src/App.js
```diff
@@ -9,6 +9,9 @@ const theme = createMuiTheme({
   palette: {
     primary: { main: "#086375", contrastText: "#ffffff" },
     secondary: { main: "#119da4", contrastText: "#ffffff" },
+    background: {
+      default: "#F2FCFF",
```
jkeary: @hhummel is this a you change or a @robb33 change?  I dont know too much about changing the overall css theme stuff using Material-UI, but seems like a strange thing to have on this branch.

frontend/test/AllQueues.test.js
```diff
@@ -0,0 +1,82 @@
+/* eslint-disable quotes */
```
jkeary: @hhummel tests look great.  would you be willing to update the front end testing wiki doc that I made with any additional information that you learned thru your process of making these tests?  https://github.com/CodeForPhilly/prevention-point/wiki/Front-End-Automated-Testing

frontend/src/stores/QueueStore.js
```diff
@@ -1,33 +1,101 @@
 import { observable, action, flow } from "mobx"
```
jkeary: @hhummel this is all looking fantastic, but the decorators are giving my linter a headache.  Take a look what I did here on the branch Robby and I are working on to get rid of it: https://github.com/CodeForPhilly/prevention-point/blob/Search_Participant_New/frontend/src/stores/ParticipantStore.js  The change was adding the `decorate` function to the bottom of the store.

frontend/src/components/AllQueues.js
```diff
@@ -36,23 +35,32 @@ const useStyles = makeStyles(theme => ({
...
 
+//Forward ref used with MUI BaseButton. Does not allow observer.
 const QueueTabContent = React.forwardRef(({ onClick, queueData }, _ref) => {
+  const queueStore = useContext(QueueStoreContext)
   const classes = useStyles()
+
+  // Add state update functionality
+
+  // Add pinging for checking if queues have updated / websockets?
+
```
jkeary: can remove these lines

frontend/src/components/QueueTable.js
```diff
@@ -58,9 +65,21 @@ function QueueTable({ queueData: { name, rows } }) {
...
       <MaterialTable
-        title={name}
+        title={queueStore.queueStats[queueData["queueData"]].name}
         className={classes.table}
-        data={rows}
+        options={{
+          search: false,
+        }}
+        data={queueStore.queues[queueData["queueData"]].map(x => ({
+          urgency: 1,
```
jkeary: are they all urgency 1?

frontend/src/routes/ServicesQueues.js
```diff
@@ -1,87 +1,141 @@
-import React, { useState } from "react"
-import PropTypes from "prop-types"
+import React from "react"
```
jkeary: @hhummel unless there is something specifically in this file you want me to look at I am not going to look at the changes here since this is stuff that Robby is working on and I have already changed this file a lot on a separate branch

frontend/src/components/QueueTable.js
```diff
@@ -48,6 +54,7 @@ function QueueTable({ queueData: { name, rows } }) {
       </IconButton>
     )
   }
+
   const SeenButton = () => {
     return (
       <IconButton>
```
jkeary: @hhummel I dont see where these things are being updated.  Shouldnt changes in status, notes, seen and urgency all have an update call made to the database?  Like calling updateQueue?  I dont see where this is happening.  My thought is that it would happen in the component, `QueueTableDropdown` but I dont see it there.  I see a handle change event in that file, but its just changing a state param in the component, and not using the Queue store at all.  Perhaps this needs to be added.  One question I have is if you think this is too much work to add to this pr and if we should ticket it out to another just to get this on master.

frontend/src/components/AllQueues.js
```diff
@@ -80,35 +88,48 @@ function QueueTab(props) {
   )
 }
 
-function AllQueues() {
+const AllQueues = observer(() => {
+  const queueStore = useContext(QueueStoreContext)
+  const queueSize = Object.keys(queueStore.queues).length
+
+  useEffect(() => {
+    for (let i = 1; i <= queueSize; i++) queueStore.updateQueue(i)
```
jkeary: so the way you have this set up is that every time you change tabs all the queues get called.  You could definitely have some logic here that doesnt do that.  I have some thoughts on how to do it, so if you dont know where to start or what to do or whatever let me know

frontend/src/components/QueueTable.js
hhummel: I update the component on render, but not other changes.  Those weren't in the original ticket

frontend/src/App.js
hhummel: The change in seed is a debugging flag i forgot to turn off when I did the Insurers PR, so I did it here.  I only affects the output when seed is run. The context in queue/views.py was to correct an error I was receiving.  The change in App.js in in response to Marcel's request that the background color be added to the theme.

frontend/src/components/AllQueues.js
hhummel: The change in refreshAuthLogic is in reponse to prettier

frontend/src/stores/QueueStore.js
```diff
@@ -1,33 +1,101 @@
...
+          60000
+      )
+    } else {
+      return 0
+    }
   }
 
-  updateQueue = flow(function*(queue) {
-    const { ok, data } = yield api.getQueue(this.queueIds["queue"])
+  updateQueue = flow(function*(queueIndex) {
```
jkeary: should this really be called updateQueue?  It seems to just get whats in the db and set it in the store.  Maybe just getQueue?  I dunno, maybe your naming convention is right if you are thinking of it from a front end perspective, it is updating the queue state param on the front... regardless we should probably figure out what the naming convention should be because I find this confusing haha

frontend/src/components/AllQueues.js
hhummel: The database calls are in parallel, so I didn't worry about it as a first cut.  I think the way to optimize the calls to the back end is to create a hash for the queue data, and use that hash in useEffect so it knows if there has to be a back end call.

frontend/test/AllQueues.test.js
hhummel: Yes

frontend/src/components/QueueTable.js
hhummel: To be honest, I'm hazy on where we are on urgency

frontend/src/components/QueueTable.js
```diff
@@ -58,9 +65,21 @@ function QueueTable({ queueData: { name, rows } }) {
   return (
```
jkeary: could be cool to add some logic for if the queues are still loading and what to display.  I started doing something akin to this on the Participant List component on my branch here: https://github.com/CodeForPhilly/prevention-point/blob/Search_Participant_New/frontend/src/components/ParticipantsList.js.  Take a look at the useEffect hook for what I did.  Right now on your branch here, it will display `no results` or something like that in the interim as its getting the queues and then get display the queue data once its fetched.  But thats not exactly correct.  It maybe cool to have a simple loading state to let the user know that the data is on the way!  Let me know if you have questions about how I implemented it.  Its also possible that later on we should add a loading state component to display a loading icon or something, but for now we can just use some basic loading text

frontend/src/components/AllQueues.js
hhummel: But I'm happy to discuss it

frontend/src/stores/QueueStore.js
hhummel: If it's confusing, then there's probably a better name.  I thought of it as something that should be called when the queues should be updated. Admittedly, it's being called more than that.  See my comment about using the hash with useEffect

frontend/src/routes/ServicesQueues.js
hhummel: I'm surprised to see getting rid of useState...

frontend/src/stores/QueueStore.js
hhummel: I remember seeing this discussed on mobx.  I wonder why I'm not having that problem

frontend/src/components/AllQueues.js
hhummel: Oh I see.  I don't have to loop over all the queues, since they aren't being displayed.

hhummel: One reason I updated the whole set is because the tabs reflect the queue size and wait time. If only one were updated the queue lengths would be stale in a hard to understand way.

frontend/src/stores/QueueStore.js
hhummel: I changed it to getQueue

frontend/src/components/AllQueues.js
hhummel: Took out comments

frontend/src/components/QueueTable.js
hhummel: I think we should pull, and then add the other changes. Marcel's PR is based on the old AllQueues, for example

frontend/src/components/QueueTable.js
hhummel: I'll see if I can get me head around it

hhummel: I'll see if I can get me head around it

hhummel: The way the mobx documentation does this is to create an observable in the store (naturally) that reflects the loading status of the queue.

https://mobx.js.org/best/actions.html#flows

frontend/src/App.js
jkeary: ok sounds good to me

frontend/src/stores/QueueStore.js
jkeary: its possible that you've turned it off on your linter or updated your linter.  But it definitely has issues with mine.  Do me a favor and in your local frontend directory do `yarn list eslint` and let me know what version of yarn you have

frontend/src/components/QueueTable.js
jkeary: so its not pulled from the backend queues table?

frontend/src/components/QueueTable.js
jkeary: ok, I made a new ticket for this

frontend/src/components/QueueTable.js
hhummel: I took a shot at doing this using the pattern from the mobx documentation, keeping track of the loading state.  I led to an infinite loop of db fetches and re-rendering, so some rethinking is in order to figure out the best way to accomplish this

frontend/src/components/QueueTable.js
hhummel: I fixed it.

frontend/src/stores/QueueStore.js
hhummel: ~/projects/prevention-point/frontend$ yarn list eslint
yarn list v1.17.3
warning Filtering by arguments is deprecated. Please use the pattern option instead.
└─ eslint@5.16.0
✨  Done in 1.11s.
~/projects/prevention-point/frontend$ yarn list eslint
yarn list v1.17.3
warning Filtering by arguments is deprecated. Please use the pattern option instead.
└─ eslint@5.16.0
✨  Done in 1.11s.

frontend/src/components/AllQueues.js
jkeary: I think its fine to get them all up front.  My only point is that there is no need to call getting all the queues once again every time you switch tabs.  Or is there?  Can you change the wait time or queue size from the queues view?

frontend/src/components/QueueTable.js
jkeary: where?

frontend/src/components/QueueTable.js
jkeary: I think we can leave this for another ticket.  I think if you are thinking of it in a global way as opposed to a component based loader like I have (using useState), then it should be a separate ticket that we can apply across the components.  I will make a ticket

frontend/src/stores/QueueStore.js
jkeary: ok yeah that's the linter I have, so is it possible that you made changes to your linter file?

frontend/src/stores/QueueStore.js
jkeary: great

frontend/src/components/QueueTable.js
hhummel: QueueTable.js

  data={queueStore.queues[queueData["queueData"]].map(x => ({
          urgency: x.urgency,
          last: x.participant.last_name,
          uid: x.participant.pp_id,
          timeElapsed: moment(x.status.created_at).format("LT"),
          status: x.status.event_type,
          service: x.service.name,
          seen: false,
          Notes: false,
        }))}

frontend/src/components/QueueTable.js
```diff
@@ -58,9 +65,21 @@ function QueueTable({ queueData: { name, rows } }) {
...
+        }}
+        data={queueStore.queues[queueData["queueData"]].map(x => ({
+          urgency: x.urgency,
+          last: x.participant.last_name,
+          uid: x.participant.pp_id,
+          timeElapsed: moment(x.status.created_at).format("LT"),
+          status: x.status.event_type,
+          service: x.service.name,
+          seen: false,
+          Notes: false,
```
marcelkooi: Should `Notes` be lowercase? And is it supposed to be boolean?

frontend/src/routes/ServicesQueues.js
```diff
@@ -1,87 +1,141 @@
...
+  hide: {
+    display: "none",
   },
   drawer: {
     width: drawerWidth,
+    flexShrink: 0,
+  },
+  drawerPaper: {
+    width: drawerWidth,
+    /* background: "#F2FCFF", */
```
marcelkooi: Should this be `background: theme.background.default` now?

frontend/src/components/QueueTable.js
hhummel: Yes, fixed

frontend/src/routes/ServicesQueues.js
hhummel: I defer to you. I assumed the purpose of moving the background color into the theme was so it would be present by default.  Is that not how it works?

hhummel: It throws an error if I do that. Cannot red property 'default' of undefined.

hhummel: Pushed the other change.  Maybe background: theme.palette.background.default?

Issue #156, opened 2019-10-16
remove commas from Urgency Enum table 


MikeyManoguerra:

MikeyManoguerra:
also, actually use the enum on the urgency visit field

MikeyManoguerra:
done

Issue #137, opened 2019-09-17
Visit-based tables link to visit rather than participant


mariekers:
Visit-based tables in back end link via visit FK rather than participant FK

Done: Visit-based tables in back end link via visit FK rather than participant FK

mariekers:
@mattusifer you want to take this one?

MikeyManoguerra:
I'll take this one

MikeyManoguerra:
#160

PR #160, opened 2019-10-26
Mikey fk med models

MikeyManoguerra:
change medical model relations to have visit FKs, update tests and seeds with enum


MikeyManoguerra:
#137 PR changes FK on medical models, closes this. 
#156 matt removed the commas, and i implemented the table

PR #178, opened 2019-11-06
Search Participant

jkeary:
This PR covers the Search Participant Sidebar, its functionality of going from the Queues View to the Results Participant List View.  It also handles the major routing issues.  And of course, tests for the Search Participant Sidebar.


frontend/src/components/ParticipantsList.js
```diff
@@ -0,0 +1,115 @@
...
+                      <Typography>{participant.pp_id} </Typography>
+                    </TableCell>
+                    <TableCell>
+                      <Typography>{participant.first_name}</Typography>
+                    </TableCell>
+                    <TableCell>
+                      <Typography>{participant.last_name}</Typography>
+                    </TableCell>
+                    <TableCell>
+                      <Typography>Gender</Typography>
```
hhummel: Why not fill in the other fields: Gender, Race, DOB? They are all available in  the api

frontend/src/components/PerDrawer.js
```diff
@@ -46,7 +46,7 @@ const useStyles = makeStyles(theme => ({
   },
   drawerPaper: {
     width: drawerWidth,
-    background: theme.palette.background.default,
+    /* background: "#F2FCFF", */
   },
   drawerHeader: {
```
hhummel: Is background gone now?  Should the commented out line be gone?

frontend/src/stores/ParticipantStore.js
```diff
@@ -1,49 +1,60 @@
...
+    this.params.last_name = data
+  }
+
+  // Getters
+  getParticipantsList = () => {
+    return toJS(this.participants)
+  }
+  getParams = () => {
+    return toJS(this.params)
   }
```
hhummel: Why are you converting these to JSON?

frontend/src/components/UserSearch.js
```diff
@@ -27,11 +51,17 @@ const UserSearch = () => {
...
+        <form className="participant-search__form" onSubmit={handleSubmit}>
           <FormGroup className="participant-search__input">
             <FormControl>
               <InputLabel htmlFor="user_id">User ID</InputLabel>
-              <Input id="user_id" name="user_id" value="" required />
+              <Input
+                id="user_id"
+                name="user_id"
+                value={participantsStore.params.userId}
+                onChange={e => participantsStore.setUserIdParam(e.target.value)}
```
crsmyz: This looks good, I'm hoping it fixes the typing/input issue we were having in this component.

frontend/src/stores/ParticipantStore.js
crsmyz: This makes sense if we are storing the data as an observable in the store variables and only converting it to JS on a get call.

frontend/src/components/PerDrawer.js
crsmyz: This looks like a leftover style that just needs to be deleted..

frontend/src/stores/ParticipantStore.js
crsmyz: It makes sense if the value in the store is an observable there by needing the JS conversion for the component

frontend/src/components/ParticipantsList.js
jkeary: @hhummel so, this list view isnt really part of this PR.  Its @angelyoung24 work that will be in a separate PR once she has completed her work on her branch.  So its probably not worth commenting on here

hhummel: Got it

frontend/src/stores/ParticipantStore.js
hhummel: Let's discuss it tonight

frontend/src/components/PerDrawer.js
jkeary: @hhummel @crsmyz I apologize, this PerDrawer file is no longer in use.  I need to just remove it.  Will do so now.

jkeary: ok removed the PerDrawer file, it is now in the MainLayout file.

frontend/src/components/UserSearch.js
jkeary: indeed it does

PR #186, opened 2019-12-04
Update queues 163

hhummel:
Modify QueueTable.js to allow updates to urgency, status, seen and notes.   Updates are recorded in the backend.  Eliminate redundant updates from backend for all queues--only the queue being displayed is updated.

Added a patch method to visits api to allow for updates.

One weird thing is I create a new note rather than updating the note.  I did this to work around mobx, but there's got to be a better solution.


frontend/src/components/AllQueues.js
```diff
@@ -87,16 +87,24 @@ function QueueTab(props) {
...
   const classes = useStyles()
   const [tabValue, setTabValue] = useState(0)
   function handleChange(event, newValue) {
     setTabValue(newValue)
   }
 
+  //Update all queues on first rendering
+  useEffect(() => {
+    for (let i = 1; i <= queueSize; i++) queueStore.getQueue(i)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
TriStarGod: You don't need to disable eslint if you extract `getQueue` from `queueStore` (ie `const { getQueue } from queueStore` ) before all `useEffect`s and add `getQueue` to `useEffect` array. The component will become more efficient. `getQueue` shouldn't change but if it did, then `useMemo` or `useCallback` can be used.

frontend/src/components/NotesDialog.js
```diff
@@ -7,16 +7,21 @@ import DialogContent from "@material-ui/core/DialogContent"
...
-  const handleSubmit = () => {
-    // Make POST request to API here to submit data
-    // Need to pass in visit ID to this component (most likely) for API
+  const handleSubmit = (queueId, visitId) => {
+    queueStore.patchVisit(queueId, visitId, { notes: participantNotes })
     toggleVisibleDialog()
   }
 
+  useEffect(() => {
+    setParticipantNotes("")
```
TriStarGod: Try the following
```
const { getNotes } = queueStore;
...
useEffect(() => {
  setParticipantNotes(getNotes(queueData, id));
}, [getNotes, queueData, id]);

```

I also don't see an `onChange` handler or value set for the `TextField` to use `participantNotes`

frontend/src/components/QueueTable.js
```diff
@@ -26,51 +26,60 @@ const useStyles = makeStyles(theme => ({
...
     { value: "LEFT", name: "Left" },
   ]
+
   const urgencyOptions = [
-    { value: 1, name: 1 },
-    { value: 2, name: 2 },
-    { value: 3, name: 3 },
-    { value: 4, name: 4 },
-    { value: 5, name: 5 },
+    { value: "ONE", name: 1 },
```
TriStarGod: Aren't the original urgency options correct since Urgency is being turned to a int? Its also necessary for urgency ordering.

frontend/src/components/QueueTableDropdown.js
```diff
@@ -1,12 +1,23 @@
-import React, { useState } from "react"
+import React, { useContext } from "react"
 import PropTypes, { oneOfType, string, number } from "prop-types"
 import MenuItem from "@material-ui/core/MenuItem"
 import Select from "@material-ui/core/Select"
+import { QueueStoreContext } from "../stores/QueueStore"
 
 const QueueTableDropdown = props => {
-  const [value, setValue] = useState(props.initialValue)
```
TriStarGod: The original useState is correct since value should reset on each re-render if you don't.

frontend/src/components/AllQueues.js
hhummel: Nice.  I added queueStore.getQueue to the useEffect array.  It's just a function, so it should not change.  I got various mobx errors if I tried either using object destructuring as you suggested, or simply assigning const getQueue = queueStore.getQueue.  Oddly, the errors were different.  Is the efficiency gain just avoiding looking up the reference (or pointer or whatever you call it in this context) associated with queueStore.getQueue?

frontend/src/components/NotesDialog.js
hhummel: Excellent.  Easy once you've explained it! I got the mobx error I described above if I used the assignment for getNotes, but works if I use queueStore.getNotes

frontend/src/components/QueueTable.js
hhummel: Are you saying these should be changed to 1, 2... here and in the back end core/models/visit.py?

frontend/src/components/QueueTableDropdown.js
hhummel: Here I have to disagree.

I had it as you suggested, but it doesn't work correctly.  If you persist state in the component with useState, then if you change a status--e.g. to "Stepped Out" in one queue--and then change tabs, you will find it's also changed in the other tabs.  

Instead you want the state persisted in the backend, and the data is refreshed from the backend when you change tabs. This is why I set value=props.initialValue

frontend/src/components/NotesDialog.js
hhummel: I got rid of the field above the input showing the previous comment, since it's now visible already. I don't think an onChange is needed. Do you?

frontend/src/components/AllQueues.js
hhummel: Because the  getQueue assignment didn't work, the linter wanted me to include queueStore in the useEffect array.  I added it and took out the sub-dependencies, (i..e. getQueue) and it looks okay,  I didn't do this from the outset because I had a vague notion that it would degrade performance, but that's not based on evidence.

TriStarGod: What type of errors? Usually you would destructure properties and then use it in all proceeding functions, otherwise, it involves multiple lookups in the object for the same properties. I believe that's eslint and airbnb's recommended guidelines. Anyway, if it works, then it works.

frontend/src/components/NotesDialog.js
TriStarGod: How do you update participantNotes without an onChange? Some lines of code may be missing in this review because I don't even see the TextField's `/>` .

frontend/src/components/QueueTable.js
TriStarGod: Yes. From what I understood, its not suppose to be words. Also, every time urgency is sorted, viewed, or saved, it would need to be converted to and from numbers unnecessarily.

frontend/src/components/QueueTableDropdown.js
TriStarGod: I'll need to play with it to get it. I probably don't understand the current context of this component.

frontend/src/components/QueueTable.js
hhummel: Got it.

PR #190, opened 2019-12-11
US85: adding participants info pages and functionality

crsmyz:


frontend/src/components/ParticipantInfo.js
```diff
@@ -0,0 +1,544 @@
...
+  const [dateOfBirth, setDateOfBirth] = React.useState("")
+  const [ppId, setPPId] = React.useState("")
+  const [race, setRace] = React.useState("")
+  const [gender, setGender] = React.useState("")
+  const [hasInsurance, setHasInsurance] = React.useState(false)
+  const [insuranceType, setInsuranceType] = React.useState("")
+  const [program, setProgram] = React.useState("")
+  const [service, setService] = React.useState("")
+  const [priority, setPriority] = React.useState("")
+  const [note, setNote] = React.useState("")
```
marcelkooi: What do you think about using one object that contains all of these properties? For instance:

```javascript
const [participant, setParticipant] = React.useState({})
```

And then from the Input component we can make the `onChange` function look like the following:

```javascript
<Input
  onChange={e => setParticipant(prevState => { ...prevState, thisProperty: e.target.value })}
  ...
/>
```

frontend/src/components/ParticipantInfo.js
```diff
@@ -0,0 +1,544 @@
...
+  const handleNoteChange = e => setNote(e.target.value)
+
+  const createStartDate = () => {
+    var now = new Date()
+    var y = now.getFullYear()
+    var m = now.getMonth() + 1
+    var d = now.getDate()
+    return (
+      "" + y + "-" + (m < 10 ? "0" : "") + m + "-" + (d < 10 ? "0" : "") + d
+    )
```
marcelkooi: Could we use a third-party library like [`date-fns`](https://date-fns.org/v2.8.1/docs/format) for this? With date-fns this could be simplified to:

```javascript
import { format } from "date-fns"

...

const createStartDate = () => {
  return format(new Date(), 'yyyy-MM-dd')
}

```

I think it'll be helpful to have something like this to handle dates in the future as well

frontend/src/components/ParticipantInfo.js
crsmyz: I love it and originally had that idea in mind but just opted for this right now for demo purposes. I'd much rather prefer one object in local state than the approach I've pushed

frontend/src/components/ParticipantInfo.js
crsmyz: Heck yea, I'll had this and re-push

hhummel: Nice

hhummel:
const changeHandler = (e, arg) => {
setParticipant(prevState => ({
...prevState,
[arg]: e.target.value,
}))
}

and then:

onChange= {e => changeHandler(e, note)}, for example?

PR #196, opened 2019-12-18
Search results page 83

angelyoung24:
Added footer and finished out my checklist.

### DIFF
--- a/frontend/src/components/ParticipantsList.js
+++ b/frontend/src/components/ParticipantsList.js
@@ -10,7 +10,9 @@ import TableHead from "@material-ui/core/TableHead"
 import TableRow from "@material-ui/core/TableRow"
 import Fab from "@material-ui/core/Fab"
 import AddIcon from "@material-ui/icons/Add"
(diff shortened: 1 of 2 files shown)


frontend/src/components/ParticipantsList.js
```diff
@@ -9,14 +10,19 @@ import TableHead from "@material-ui/core/TableHead"
...
+import Grid from "@material-ui/core/Grid"
+import PersonAddIcon from "@material-ui/icons/PersonAdd"
+import Paper from "@material-ui/core/Paper"
 import { observer } from "mobx-react-lite"
 import { Link } from "react-router-dom"
 
 const ParticipantsList = observer(() => {
   const rootStore = useContext(rootStoreContext)
   const participantsStore = rootStore.ParticipantStore
-  const [isLoading, setIsLoading] = useState(false)
```
mariekers: this should be re-added

mariekers:
@angelyoung24 - can you look at the failed check reasons and fix this branch so we can merge? Will this fix the search results to they actually filter to the search input?

PR #197, opened 2019-12-18
Filter users by search params

### LINKED ISSUE #192: Fixing "Participant Search" page to show filtered results.
Currently the "ParticipantList" component does not show filtered results for current participants. 

As a user I should be able to search for a participant by ID or by first name and last name and see only results that fit my parameters.

Example:

Search by FL3498.   =>  result, I should see only participants with that ID

Search by Darius Brown => result, I should see only participants with that first and last name.

marcelkooi:
Closes #192


crsmyz:
Reviewing

PR #198, opened 2019-12-19
US bug fix for participant info page

crsmyz:


frontend/src/components/ParticipantInfo.js
```diff
@@ -110,12 +127,14 @@ const ParticipantInfo = observer(() => {
...
-    }))
-  }
+  // const changeParticipant = (e, args) => {
+  //   // eslint-disable-next-line no-console
+  //   console.log(args)
+  //   setParticipant(participant => ({
+  //     ...participant,
+  //     args: e.target.value,
+  //   }))
+  // }
```
marcelkooi: Can we remove this since we're not using it?

frontend/src/components/ParticipantInfo.js
```diff
@@ -316,7 +366,12 @@ const ParticipantInfo = observer(() => {
...
-                        onChange={e => changeParticipant(e, "hasInsurance")}
+                        onChange={e =>
+                          setParticipant({
+                            ...participant,
+                            hasInsurance: e.target.value,
+                          })
+                        }
                         style={{ display: "inline" }}
                       >
                         <FormControlLabel
```
marcelkooi: Can we change the values for these radio buttons to `value={true}` and `value={false}` since that's how the Participant object is being populated?

PR #203, opened 2020-01-16
US200: Fixing sort on QTable for time and Urgency

crsmyz:


frontend/src/components/QueueTable.js
```diff
@@ -103,10 +103,18 @@ const QueueTable = observer(queueData => {
...
           { title: "Last", field: "last" },
           { title: "UID", field: "uid" },
-          { title: "Time", field: "timeElapsed" },
+          {
+            title: "Time",
+            field: "timeElapsed",
+            // eslint-disable-next-line no-console
+            customSort: (a, b) =>
+              new Date("1970/01/01 " + a.timeElapsed) -
+              new Date("1970/01/01 " + b.timeElapsed),
```
marcelkooi: What kind of value do we expect `timeElapsed` to be?

crsmyz: @marcelkooi timeElapsed should be the string of the hour and minutes, the format looks like "hh:mm am/pm"

PR #206, opened 2020-01-21
UStbd: Hiding search bar drawer on login page

crsmyz:


crsmyz:
> Good. Should we refactor to get rid of the hard coded nav bar?

I think the hard coded template is good for now, we can modularize it more moving forward though.

PR #216, opened 2020-02-02
Program availability toydata

quinn-dougherty:
- generates fake data for the table `core_programavailability` / `models.ProgramAvailability` 

small things: 
- my ide generated `.mypy_cache/` so i `.gitignore`'d it. 
- the faker provider `MonFri` isn't used in the last commit, but I left it in there because it might come in handy later. 
- I tried to have the same opinions as the rest of the document, one difference is I avoided one-letter var names.

### DIFF
--- a/core/management/commands/seed.py
+++ b/core/management/commands/seed.py
@@ -1,4 +1,4 @@
-import random, pytz
+import random, pytz, datetime
 
 
 from django.utils import timezone
@@ -20,6 +20,7 @@ from core.models import (
     Visit,
     ProgramServiceMap,
     UrgencyLevel,
+    ProgramAvailability
 )
 from core.permissions import CASE_MANAGER, FRONT_DESK, ADMIN
 from core.models.front_desk_event import FrontDeskEventType, FrontDeskEvent
@@ -32,6 +33,7 @@ DEFAULT_GROUPS = [FRONT_DESK, CASE_MANAGER, ADMIN]
 DEFAULT_NUMBER_PARTICIPANTS = 1000
 DEFAULT_NUMBER_VISITS = 100
 DEFAULT_NUMBER_INSURERS = 10
+DEFAULT_NUMBER_PROGRAM_AVAILABILITIES = 10
 
 # Cribbed from prevpoint-backend 2 July 2019 Marieke
 DEFAULT_PROGRAMS = {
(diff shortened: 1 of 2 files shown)


quinn-dougherty:
@hhummel

core/management/commands/seed.py
```diff
@@ -32,6 +33,7 @@
 DEFAULT_NUMBER_PARTICIPANTS = 1000
 DEFAULT_NUMBER_VISITS = 100
 DEFAULT_NUMBER_INSURERS = 10
+DEFAULT_NUMBER_PROGRAM_AVAILABILITIES = 10
```
hhummel: There are about 8 programs and 5 weekdays, so 8 * 5 = 40.  But you only have 10 PROGRAM_AVAILABILITIES, so hardly any programs will be open on a given day.  You could simply increase the number, but then you might have overlapping and conflicting availabilities, e.g. CM available on Friday 8am-3pm but also 9am-4pm.  I think a more natural way to do it would be to cycle through the programs and days of the week and either have random hours available, or none at all.

core/management/commands/seed.py
```diff
@@ -298,6 +327,29 @@ def create_event(visit, type):
     f.full_clean()
     f.save()
 
+
+def create_program_availability(output=True):
+    """create program availability"""
+    programs = Program.objects.all()
+    for _ in range(DEFAULT_NUMBER_PROGRAM_AVAILABILITIES):
```
hhummel: So I would do:
for program in programs:
    for day in ["Mon", "Tues", ...]:
        randomly hours, or randomly no hours or randomly more than one set of hours, (like a lunch break)
        create and save the object.

quinn-dougherty: briefly while i'm working on this, Marieke on slack said to include Saturday and Sunday

quinn-dougherty: This change completed https://github.com/CodeForPhilly/prevention-point/pull/216/commits/c0ce037ecd53e83aca5d1f9857803de3dd8929fc

core/management/commands/seed.py
hhummel: Does this parameter do anything now? I think you can cut it.

core/management/commands/seed.py
```diff
@@ -298,6 +327,52 @@ def create_event(visit, type):
...
+                    availability_two.day_of_week = day
+                    availability_two.start_time, availability_two.end_time = fake.availability_window(
+                        start_hour=window_one_end+1, end_hour=23)
+                    availability_two.full_clean()
+                    availability_two.save()
+
+                    if output:
+                        print(
+                            "Created program availability: {} {} {} {}".format(
+                                availability_two.program,
```
hhummel: When you enable output, the program labeling isn't very helpful.  How about making these like "availability_two.program.name".  In django you can follow the fields in the foreign key object like program is here.  Then it will show labels like "CM" for case management, for example.

core/management/commands/seed.py
```diff
@@ -298,6 +327,52 @@ def create_event(visit, type):
...
+                if output:
+                    print(
+                        "Created program availability: {} {} {} {}".format(
+                            availability.program,
+                            availability.day_of_week,
+                            availability.start_time,
+                            availability.end_time
+                        )
+                    )
+
```
hhummel: I thought the output from this approach was good, (modulo the comment above), but the code was a little awkward.  What do you think about separating the output like:
if output:
    availabilities = ProgramAvailability.objects.all()
    for a in availabilites:
        print(f"Created {a.program.name} {a.day_of_week} {a.start_time} {a.end_time}")

If the output needs to be sorted to be understandable, you can do that using django ORM after the objects.all()

Just a thought

quinn-dougherty: pushed

mariekers:
@swearingbird mind giving this a once-over?

mariekers:
What does this need to merge? @MikeyManoguerra @hhummel

Issue #212, opened 2020-01-28
dockerize front end 


mariekers:
Project needs to be fully dockerized

Done: front end component added to docker-compose

**Required to host in linode

ThrowsException:
I might be able to help with this. What would the container actually need to provide to be usable by linode? Something like an nginx container with ports exposed serving the frontend assets or just a container with the assets built in folders?

mariekers:
@ThrowsException I'm going to direct this to @themightychris

mariekers:
the main issue here is to add the front end to the docker-compose file as a second service

themightychris:
Please expose an HTTP port from the container. CfP ops will handle putting SSL in front of it

hhummel:
@ThrowsException are you working on this?

ThrowsException:
@hhummel hey sorry Ive been away for a few days I was hoping to jump on this today. Shouldnt be to long to get a working nginx container up

mariekers:
@ThrowsException is this done?

ThrowsException:
@mariekers Ive merged in @themightychris changes to the PR awaiting review.

themightychris:
It was good to go in my book pending that merge, just marked the PR approved, are we waiting on anyone else?

ThrowsException:
@themightychris I don't think so. I think its just been me you and @mariekers. Want me to merge?

themightychris:
![giphy (1)](https://user-images.githubusercontent.com/458494/75407285-98575400-58e0-11ea-8fbc-2f91656e615d.gif)

PR #218, opened 2020-02-05
Dockerize client application

ThrowsException:
#212 

**What**
Attempt at building a container for the frontend that can be deployed. 
Builds frontend application into a container running nginx exposed on default port 80
Docker compose file exposes build container on port 8080 
`docker-compose up -d client` makes the frontend available.


ThrowsException:
@themightychris I've updated this PR with the discussions on #223 

The way this works is the nginx container hosts the client app still but proxies requests to `/api` to the container running the backend. Thankfully all the endpoints the frontend needs are at `/api` so that makes that pretty simple. This setup should work using both `yarn dev` if you're working on front end or running the entire composer setup with `docker-compose up -d --build` 

Couple issues I **DONT** like about how I have this set up. 
This is a not what I would consider ready for prod but as you said maybe this can just be a template to move it to whatever platform you need to deploy it. Whoever the reviewer is here you'll see I needed to turn off some of the interactivity for some of the django setup to work in the entrypoint script. A lot of these commands as well such as `seed` and `createsuperuser` only seem useful right now from a development perspective to be automatically run while bootstrapping the project. I assume `seed` would not be run in production and `createsuperuser` and `migrate` would probably be better handled with manual intervention on whatever server/container is running unless someone has a better way to dockerize these commands. Could possibly do it with environment variables but a missing environment variable with the change to blow away a database by accident makes me nervous. 

Comments welcome

themightychris:
@ThrowsException could you outline what you'd flag as making this unsuitable for production? Or did you mostly cover that re: `seed` / `createsuperuser` / `migrate` ?

themightychris:
I've found some good results following the `Script To Rule Them All` pattern for developer workflows: https://github.com/github/scripts-to-rule-them-all

With this pattern, we might move all initialization out of the docker entrypoint so we can get closer to having the same container image for dev/prod. A `./script/server` command could handle launching the container with any overrides/envs a dev would want while a `./script/setup` command would handle running any commands against the container that a dev creating a fresh environment would want

ThrowsException:
Ill try to apply the script pattern to it but yes the parts you mentioned `seed` `superuser` `migrate` are the parts I would consider not prod ready. Theres also nothing like dotenv to control links int he client application. I'll try refining what I have maybe starting with a dev and prod script and see if we like that approach more.

themightychris:
Ideally we could avoid dotenv and place needed envs in `docker-compose.yml` as baselines or in `./script/server` as developer environment overrides, with default defined in the `Dockerfile` as much as possible

IME dotenv gets messy going into cloud-native deployment. The `docker-compose.yml` can include loading a local env-file for local overrides, but anything that would be standard for any dev setup could go in `./script/server`. Any application-level loading of .env files can be problematic when moving into environments where we actually need everything to be environmental, and running under docker-compose gives us a all the tools we need to do that outside the application code

ThrowsException:
@themightychris I updated the setup some more and I like this better. First big change replace the django dev server in the container with uwsgi which is going to be better for a deployed production container but can also be started in a development environment. I also introduced the `scripts` folder with a `bootstrap.sh` to start the fill environment and a `develop.sh` to seed all the test and dev data. At this point it would probably be a matter of adding whatever scripts you would need to run in a production environment and adding whatver environment variable need to be overridden and added to the `core/settings/production` Django file. Im leaving that as an exercise for ops when it comes times to deploy but I think this leaves the Django container in a much better place.

themightychris:
@ThrowsException any opposition to using the standard script names? That means no `.sh` suffix and renaming develop to setup (these scripts are inherently for developers)

There's some thought we might map their presence in project repos as part of health indicators if it doesn't prove too awkward for different stacks to implement the interface

ThrowsException:
@themightychris Nope. updated scripts and readmes.

mariekers:
@ThrowsException @themightychris what's the status on this PR - can it be merged? Are we waiting on the back end dockerization? 

Prevention Point is very eager to play with the system - I'm getting calls and emails every other day.  They're excited to actually have something, so sorry if I'm coming off as demanding!

ThrowsException:
@mariekers I think I was just waiting on the final ok. AFAIK this is good to go.

themightychris:
Merge this PR first: https://github.com/ThrowsException/prevention-point/pull/1

PR #219, opened 2020-02-07
US215 adding navigation from q table to participant info

crsmyz:


frontend/src/components/QueueTable.js
```diff
@@ -91,6 +93,13 @@ const QueueTable = observer(queueData => {
...
           id: x.id,
         }))}
+        actions={[
+          {
+            icon: "assignmentIndIcon",
+            tooltip: "Edit Participant",
+            onClick: () => history.push("/participantInfo"),
+          },
+        ]}
         columns={[
```
hhummel: This all looks good, but in the form 5 is highest priority, but in the default sorting 1 is highest

hhummel: Could you fix that?

hhummel:
Stub for from in QueueTable

Issue #214, opened 2020-01-29
temporarily "freeze" queue


mariekers:
As a provider, I want to be able to temporarily stop the front desk from adding new people to a queue. 

Done: add button/check box to queue page to stop *that* queue - stop button should be relative to active queue tab

ThrowsException:
Unsure of the requirements here. To me it seems people are added to queues by visiting the /participantInfo endpoint and filling out the information and specifying the queue (Program?). 

Seems like the update here if im reacding it right is the backend needs to know which queues are "active" and only display programs in the drop down that a hold has not been placed which would require some backend work as well.

mariekers:
Excellent question, @ThrowsException 

Providers need to be able to tell the person at the front desk to temporarily stop adding participants to the queue, so they need the ability to 
 - [ ] mark a program as temporarily frozen from the queue page
 - [ ] program is unavailable in dropdown on participant detail page
 - [ ] provider can "unfreeze" a program from the queue page

This probably has front and back end work - I'd love to hear your thoughts on the best way to implement this!

MikeyManoguerra:
This is finished!

PR #224, opened 2020-02-16
Freeze Queue

ThrowsException:
#214 

Makes the program not selectable from the participant info form if the program has been marked as frozen from the queue


frontend/src/components/AllQueues.js
```diff
@@ -128,6 +142,17 @@ const AllQueues = observer(() => {
         </Tabs>
       </AppBar>
       <QueueTable queueData={tabValue + 1} />
+
+      <div
+        style={{
```
MikeyManoguerra: best to avoid inline styles, can you create a scss/queue file and add this there

ThrowsException:
@MikeyManoguerra thanks for the review. I revamped this a little bit. Instead of just blindly flipping the state when its clicked the button will wait for a valid server response. I also went a fixed the AllQueue.test.js file since it was failing due to some other PR's. Also added a test for the button. Probably want to give this a once over again.

MikeyManoguerra:
Looks good!
Its ok to merge, i believe convention still is once you get approved you merge your own changes.

Some things outside the scope of this ticket we'll need to think about: 
- the difference between "frozen" and "closed". 
- A daily cycle : does "is_ frozen" reset to false every day?  This can fold into some sort of daily reset routine, That should be discussed in full because there are definitely other undecided lifecycle type things to consider

ThrowsException:
@MikeyManoguerra Yea I noticed the is closed. if its decided later to use that instead it should be easy to switch over. ill get the branch updated and merged.

ThrowsException:
I guess by rebasing it invalidates the review. Sorry mind approving again and ill get this merged.

MikeyManoguerra:
@ThrowsException they're moving too fast! message me on slack and ill see this quicker

PR #233, opened 2020-03-02
Serve static assets in docker container

ThrowsException:
Build static assets and serve from uwsgi

### DIFF
--- a/script/setup
+++ b/script/setup
@@ -15,10 +15,6 @@ echo "==> Initializing and seeding database…"
 docker-compose exec app python manage.py migrate
 docker-compose exec app python manage.py seed
 
-echo 
-echo "==> Initialize static assets"
-docker-compose exec app python manage.py collectstatic
-
 echo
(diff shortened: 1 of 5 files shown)


script/setup
```diff
@@ -15,6 +15,10 @@ echo "==> Initializing and seeding database…"
 docker-compose exec app python manage.py migrate
 docker-compose exec app python manage.py seed
 
+echo 
+echo "==> Initialize static assets"
+docker-compose exec app python manage.py collectstatic
+
```
themightychris: This should probably go in update and server

Does this need to be re-run after code/branch changes? Does it need to be run once before starting a server?

ThrowsException: Should have clarified more. So this effects assets in the /admin and /swagger section of the django backend. 

Maybe a better way to handle this is to run `collectstatic` while building the container and dump the assets in. Thoughts?

ThrowsException: Decided this was probably better in the server and update scripts.

PR #234, opened 2020-03-03
US228: Login form enter key bug fix

crsmyz:


ThrowsException:
The `form` should probably have an `onSubmit` action as well.

crsmyz:
> Looking at how something like formik does this. I think its better to move the login handler to the `onSubmit` of the form and do the change your doing here by changing it to `type="submit"`

The current push should reflect what your looking for.

Issue #238, opened 2020-03-09
Add Metabase to docker


mariekers:
As a user, I would like to be able to build metabase reports against the fake data created in the back end of the MVP.

Done: 
Add metabase as a service to docker file (maybe? open to other ideas)

We have a basic version of [metabase](https://github.com/CodeForPhilly/prevpoint-metabase), but it's using its own postgres instance - metabase allows us to set up a connection to the actual db within the metabase UI so the postgres in the metabase container may not be necessary.

ThrowsException:
Is there a current metabase instance running somewhere? Why can't that connect to the prevention point database?

mariekers:
going to host metabase separately

PR #239, opened 2020-03-09
Add metabase docker container

ThrowsException:
Adds local metabase container #238

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,20 @@ This project implements the [Scripts To Rule Them All](https://github.com/github
 * `script/setup` — Run after `script/server` to initialize local database with seed data
 * `script/update` — Run after changing code or switching branches to refresh server
 
+## Metabase
+
+The Metabase containers can be configured to the postgres database by using there settings aftering creating the containers. 
+| Setting           | Value             |
+| ----------------- | ----------------- |
+| Name              | (Your preference) |
+| Host              | db                |
+| Port              | 5432 (default)    |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -76,6 +76,20 @@ This project implements the [Scripts To Rule Them All](https://github.com/github
...
+
+The Metabase containers can be configured to the postgres database by using there settings aftering creating the containers. 
+| Setting           | Value             |
+| ----------------- | ----------------- |
+| Name              | (Your preference) |
+| Host              | db                |
+| Port              | 5432 (default)    |
+| Database name     | postgres          |
+| Database username | (leave blank)     |
+| Database passowrd | (leave blank)     |
```
themightychris: Can this config get passed into the metabase container via env from the docker-compose file?

ThrowsException: I did look around their docs I didn't see a way to have a default config with either a mounted config file or env variables. The only other thing I could think of is setting up an initial connection and checking in the `.db` files it creates and using some scripting up some kind of startup but I wasn't sure if that was worth the effort for this

mariekers:
@MooseQuest this was to make it so the app and metabase were spun up at the same time

keegan2149:
I’m not getting an error message during the docket compose. I can just add the environment vars for Postgres. I feel like I’m missing something though.

MooseQuest:
Lets setup sometime to review this,maybe tomorrow?

PR #254, opened 2020-03-23
US253: Refactoring ParticipantInfo component

crsmyz:


frontend/src/stores/ParticipantStore.js
```diff
@@ -87,7 +93,7 @@ export class ParticipantStore {
     this.insurers = data
   }
   @action setPrograms = data => {
-    this.programs = data
+    this.programs = data.filter(item => !item.is_frozen)
```
thatzacdavis: This looks like a good defensive check, but is there a way to try to enforce this in the UI as well?

crsmyz: Could be? honestly I'd need to double check that functionality for freezing, mostly because I didn't directly work on it. The above code just got moved from the component to the store.

crsmyz: @thatzacdavis when you mean enforce this in the UI what specifically did you have in mind?

thatzacdavis: Still learning my way around the app, but I was thinking a read-only mode on these if they are presented in the UI.

crsmyz: MobX has a strict mode that makes it so only methods marked as "@actions" can modify state, this is the closest analog for MobX I can think off. But if your talking about the component itself then I would think its a matter of immutability, right?

thatzacdavis: Yes I think it is. Probably something for another ticket.

frontend/src/components/ParticipantInfo.js
```diff
@@ -1,3 +1,10 @@
+// TODO:
```
MikeyManoguerra, line 1: I very much agree with these statements

Issue #167, opened 2019-10-29
Verify User's Access to Program


mariekers:
Django can now create users with permissions to specific objects in the back end. 
[Wiki](https://code.djangoproject.com/wiki/RowLevelPermissionsDeveloper)

Users that require permissions to specific objects
Front Desk 
   - [ ] Participant
   - [ ] Visit

External provider 
   - [ ] Participant
   - [ ] visit
   - [ ] program specific objects

Related to #165

mattusifer:
This is more about adding authorization around programs - we need to check in each endpoint whether the given user has access to the program that they are trying to access, 403 otherwise

mattusifer:
@mariekers how can we tell if a user is an "external" user? is that something that should be added to the users table?

MikeyManoguerra:
@mattusifer , @mariekers , i was a little more specific on the other ticket that talks about this. Can we add another group "external providers" and then if they are check the mapping table you describe?

MikeyManoguerra:
More info is needed before this issue is tackled, because it will require a permissions system overhaul.

Front Desk:
- Can they see all participant objects? can they see any medical table data?

Internal Provider:
- Can every type of Internal Provider have access to every Participant's entire medical data

External Provider:
- Can Participants fall into groups?  or does a provider type have to have a completely custom patient set?
- Can  a certain type of external provider ( Heart Doctor For example )  have access to all the heart patients, or only a custom subset ( can e split the difference somehow)

MikeyManoguerra:
Here are the notes relevant info  from #165

Front desk users can access everything used by the MVP

External providers should be able to access everything the front desk accesses + their program's data

Internal providers can access everything (minus the admin functionalities)

Admins can access everything

MikeyManoguerra:
the logical leap this needs to take is a dynamic permission. PPP needs to be able to add users to groups, and those groups need to have dynamic model level permission based upon their role.

PR #255, opened 2020-03-28
permissions upgrade

MikeyManoguerra:
This PR removes custom permission class and implements DjangoModelPermissions from DRF. 
It still uses groups to seperate users, but now groups can have programmatically assigned permissons


MikeyManoguerra:
#167

MikeyManoguerra:
Sorry this touches so many files.

 the biggest change is in `core/permissions.py` 
the permissions are seeded by `commands/users_and_groups.py`

PR #260, opened 2020-04-08
US257: Load error on edit participant from queue

crsmyz:


crsmyz:
> i assume there was disjointed stack issue with setting data in the useeffect in stead of the store?
> looks good!

Yea it looked like the useEffect was firing the loop before the program list was coming back causing the error. I've just moved that logic to the store now with the same checks as before.

PR #265, opened 2020-04-14
Demo deploy

machow:
This PR allows the app to run on hosts that are not localhost. There were two main issues: Django is not set to accept any hosts, and in prod it's both accessible via nginx (good) and a port straight to the app server (bad? probably not a big issue).

Changes:

* set django `ALLOW_HOSTS = ['*']`, remove host-exposed backend port.
* comment out dev settings in docker-compose.yml (eg nginx config bind mount; exposing backend port on host since it is accessed via nginx)

Note: I also commented out parts of nginx while trying to figure out how to get non localhost hosts working. Could likely uncomment again!

```
        #proxy_set_header   X-Real-IP $remote_addr;
        #proxy_set_header   X-Forwarded-For $proxy_add_x_forwarded_for;
        #proxy_set_header   X-Forwarded-Host $server_name;
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,8 +9,8 @@ services:
       - data:/var/lib/postgresql/data/
   app:
     build: .
-    ports:
-      - "8000:8000"
+    expose:
+      - "8000"
     depends_on:
       - db
     environment:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,7 +21,7 @@ services:
       - "8080:80"
     depends_on:
       - app
-    volumes:
-      - ./frontend/nginx.conf:/etc/nginx/conf.d/default.conf:ro
+#   volumes:
+#     - ./frontend/nginx.conf:/etc/nginx/conf.d/default.conf:ro
 volumes:
   data:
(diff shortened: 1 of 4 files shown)


MikeyManoguerra:
This PR reminded me that there are some django specific production things that should be done.

 however, does this apply when it is inside a docker container? 
- debug should be set to false in settings/poduction
- secret key should be an env variable?

docker-compose.yml
```diff
@@ -9,8 +9,8 @@ services:
       - data:/var/lib/postgresql/data/
   app:
     build: .
-    ports:
-      - "8000:8000"
+    expose:
+      - "8000"
```
ThrowsException, line 13: this still exposes port 8000 to the host just now on a random port.

frontend/nginx.conf
```diff
@@ -18,8 +18,8 @@ server {
         proxy_pass         http://docker-app;
         proxy_redirect     off;
         proxy_set_header   Host $host;
-        proxy_set_header   X-Real-IP $remote_addr;
-        proxy_set_header   X-Forwarded-For $proxy_add_x_forwarded_for;
-        proxy_set_header   X-Forwarded-Host $server_name;
+        #proxy_set_header   X-Real-IP $remote_addr;
```
ThrowsException, line 21: if planning on using in production these should be uncommented. Proxy should set these for logging and tracking purposes.

PR #282, opened 2020-05-01
Cleanup some of the AllQueue code

ThrowsException:


ThrowsException:
@MikeyManoguerra Not aware of the issue but I didn't see any console errors while testing. Is there an existing issue open ?

PR #284, opened 2020-05-03
PR269: wrapping MUI input with PrevPoint Comp

crsmyz:
This PR wraps the MUI input component in a PrevPointInput wrapper that manages props with:

id,
name,
val,
changeFunc
changeFunc2

This allows only 1 import from MUI and help performance not to mention allows more modular and custom use of these components.

### DIFF
--- a/frontend/src/components/ParticipantInfo.js
+++ b/frontend/src/components/ParticipantInfo.js
@@ -12,7 +12,6 @@ import "../scss/participant-search.scss"
 import FormGroup from "@material-ui/core/FormGroup"
 import FormControl from "@material-ui/core/FormControl"
 import InputLabel from "@material-ui/core/InputLabel"
-import Input from "@material-ui/core/Input"
 import Typography from "@material-ui/core/Typography"
 import Grid from "@material-ui/core/Grid"
 import TextField from "@material-ui/core/TextField"
@@ -28,6 +27,7 @@ import Button from "@material-ui/core/Button"
 import { observer } from "mobx-react-lite"
 import { useHistory } from "react-router-dom"
 import { autorun, toJS } from "mobx"
+import PrevPointInput from "./PrevPointInput"
 
 const useStyles = makeStyles(theme => ({
   paper: {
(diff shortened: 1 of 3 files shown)


frontend/src/components/ParticipantInfo.js
```diff
@@ -229,17 +223,12 @@ const ParticipantInfo = observer(() => {
...
-                      onChange={e => {
-                        participantStore.setPPId(e.target.value)
-                        participantStore.setLastFourSSN(
-                          +e.target.value.substr(2)
-                        )
-                      }}
-                      required
+                      val={existingParticipant.pp_id}
+                      changeFunc={participantStore.setPPId}
+                      changeFunc2={participantStore.setLastFourSSN}
```
ThrowsException, line 231: I'm -1 on `changeFunc2` the parent component should only pass in a single function that handles updating

crsmyz: Yea I overlooked this a bit, the reason for this is because of this code:
```onChange={e => {
      participantStore.setPPId(e.target.value)
      participantStore.setLastFourSSN(+e.target.value.substr(2)
)```
But I can just combine those two into one action from the MobX side.

Comment (ThrowsException) [reply frontend/src/components/ParticipantInfo.js:231]: if it makes sense the mobx fine but does it make more sense for the component to just pass in a function like 

```
const updateUUID = (e) => {
    participantStore.setPPId(e.target.value)
     participantStore.setLastFourSSN(+e.target.value.substr(2))
}

<PrevPointInput changeFunc={updateUUID} />
```

File: frontend/src/components/PrevPointInput.js [hunk 2]
```diff
@@ -0,0 +1,32 @@
...
+import PropTypes from "prop-types"
+import Input from "@material-ui/core/Input"
+
+const PrevPointInput = props => {
+  return (
+    <Input
+      id={props.id}
+      name={props.name}
+      value={props.val}
+      onChange={e => props.changeFunc(e.target.value)}
```
ThrowsException, line 19: Ill approve as is but I see down the line components being interested in more than just the value from the action.

crsmyz: Should allow for just the event to be emitted now. Check out what I pushed and see if it lines up,

Issue #269, opened 2020-04-26
Make reusable wrappers


mariekers:
- take reused components identified in #268 and turn them into reusable wrappers

Goal: Take Material-UI components used throughout the app and making them DRY.

Wrapping components should take in props
Material-UI components slated for wrapping include but not limited too:
@crsmyz  
`<h1-h5>`
`<InputLabel>`
`<Input>`

@MikeyManoguerra 
`<button>`
`<FormGroup>`
`<Typography>`
etc.

More Material-UI components may be eligible for wrapping than the ones listed above so always look for opportunities to wrap.

MikeyManoguerra:
This round of this concept is finished. specific wrappers can get their own tickets

PR #286, opened 2020-05-04
Wrappers for button and typography

MikeyManoguerra:
This creates PrevPointButton, and PrevPoint typography components. 
this allows out typography to be configured in one place and consistent across the app.

i also added some `pipenv run` scripts

this can wait to merge until the other pending PRs for today are merged

#269


frontend/src/components/PrevPointButton.js
```diff
@@ -0,0 +1,55 @@
...
+  },
+}))
+
+const PrevPointButton = props => {
+  // disabled and large are booleans. can add a class where button is imported
+  const classes = useStyles()
+  const { type, children, disabled, large, color, onClick, className } = props
+
+  return (
+    <Button
```
ThrowsException, line 19: I left the same comment I left on @crsmyz for the `<PrevPointInput>` component. if were wrapping components like this then it should spread the props out over the wrapped component in order to maintain compatibility and the override our specific implements details 

`<Input {...props} custom1={value} custom2={value} />`

MikeyManoguerra: The idea is to actually remove customise-ability  from the Material UI button, at least in terms of styling. The UI for this app is pretty small, and new devs should not be styling a new button, but using the same button across the app.

 If I spread the props, wont that allow all materialUI/Button props to be available again?  is there a better way to limit variability?

MikeyManoguerra: same thing with the Typography components

frontend/devServer.js
```diff
@@ -4,6 +4,10 @@ const ParcelProxyServer = require("parcel-proxy-server")
 // https://github.com/charlieduong94/parcel-proxy-server
 const server = new ParcelProxyServer({
   entryPoint: "./public/index.html",
+  // makes rebuild in dev less taxing V
+  // parcelOptions: {
+  //   sourceMaps: false,
+  // },
```
crsmyz, line 10: Is this an option we can keep around or should we nix it completely

MikeyManoguerra: Im still trying to get my head around the parcel module builder. when i put that in the hot module reload was making my computer go crazy, and this was a suggested fix. I didn't want to leave it live bc i don't know the side-effects of it. I can remove.

PR #291, opened 2020-05-09
Fix buttonbase warning

ThrowsException:
Fix warnings coming from buttonbase tab due to not displaying children. 

https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/ButtonBase/ButtonBase.js#L289-L303

  ```
const enableTouchRipple = mountedState && !disableRipple && !disabled;


  if (process.env.NODE_ENV !== 'production') {
    // eslint-disable-next-line react-hooks/rules-of-hooks
    React.useEffect(() => {
      if (enableTouchRipple && !rippleRef.current) {
        console.error(
          [
            'Material-UI: The `component` prop provided to ButtonBase is invalid.',
            'Please make sure the children prop is rendered in this custom component.',
          ].join('\n'),
        );
      }
    }, [enableTouchRipple]);
  }
```

### DIFF
--- a/frontend/src/components/QueueComponent/QueueTab/QueueTab.js
+++ b/frontend/src/components/QueueComponent/QueueTab/QueueTab.js
@@ -4,9 +4,10 @@ import QueueTabContent from "./../QueueTabContent/QueueTabContent"
 
 const QueueTab = props => (
   <Tab
+    {...props}
+    disableRipple
     component={QueueTabContent}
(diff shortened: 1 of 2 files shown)


ThrowsException:
@MikeyManoguerra if we ever did need the ripple effect the custom tab we have needs to render `{props.children}` and it also messes with the styling because it adds two divs that make up the ripple effect. just notes for the future

PR #293, opened 2020-05-11
Speed up the AllQueue page and requsts to queues

ThrowsException:
Tried to work around some of the slowness of the ModelSerializer and also consolidate some of the database trips. Biggest improvements were loading the related entities and not performing the front desk events query in a loop and also moving it to use `.values(...)` and skipping the serializer altogether. Yields some decent improvements locally. 
Before:
master
After:
performance

### DIFF
--- a/core/queue/views.py
+++ b/core/queue/views.py
@@ -2,7 +2,6 @@ from rest_framework import viewsets, status
 from rest_framework.response import Response
 from core.models import Visit, FrontDeskEvent, FrontDeskEventType
 from core.visits.serializer import VisitWithPopulationSerializer
-from core.front_desk_events.serializer import FrontDeskEventForQueueSerializer
 from core.permissions import FRONT_DESK, CASE_MANAGER, ADMIN, HasGroupPermission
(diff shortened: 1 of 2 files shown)


ThrowsException:
@MikeyManoguerra I moved the sorting to the database rather than code. No need for distcint since `IN` should bring back distinct objects.

MikeyManoguerra:
My thought was if you `distinct(visit_id) ` then the there would only be the most recent front desk event for each visit.(because you can use distinct on column it seems). It's trivial though.

ThrowsException:
@MikeyManoguerra I tried distint in the query but I wasn't getting the same results. I think I would've had to do some sort of group by to get the latest event but this should accomplish the same thing.

MikeyManoguerra:
Ok, and to be fair I have never used that method before so it might behave differently than i am reading into it.

ThrowsException:
all good I didn't spend a crazy amount of time trying to debug it either ha. But still good suggestion since getting rid of that sort and moving it into the DB was a better idea anyway.

PR #296, opened 2020-05-12
wip: seperate ParticipantInfo form into form section components

MikeyManoguerra:


ThrowsException:
One general comment on this is rather than just simply copying out things into seperate components is to try and make them pure components that get their properties from a parent. e.g the participant form IMO shouldn't need the full rootStoreContext it should be fed props from some parent component and any thing that needs to the context should be the responsibility of the parent component not the reusable component.  

To me just splitting out the components this way only adds another level of indirection plus the overhead of needing to import the context in two places instead of one. 

Maybe I'm wrong here or maybe there needs to be some type of HoC/Hook that has access to the context that we can wrap components in to give them access to functions something like `withParticipantStore(<Component />)` or something.  just my two cents

MikeyManoguerra:
That makes sense to me. Im not sure the best way to handle the context/mobx interplay

https://mobx.js.org/refguide/observer-component.html#using-context-to-pass-observables-around
https://mobx.js.org/refguide/observer-component.html#when-to-apply-observer

Something to keep in mind is that this ParticipantInfo component is going to be the participant form, plus, the new visit form,  past visits list, single visit info, medical data form, etc. there is a design example of this somewhere. Id like to know what you think about how to best handle this component considering how important its supposed to be eventually

ThrowsException:
@MikeyManoguerra yea its one of the reasons I don't like about mobx. I've always subscribed to the pure functional way of react. Components should be no smarter than they need to a rely solely on passed props or internal state most of the time with a few high level components taking care of cross talk. Mobx seems to want the approach of just make everything observers and give every component access to everything all the time. I haven't spent much learning the ins and outs of mobx but thats what is seems. 

We can talk more in the #prevpoint-dev room if you want. if this were me id have the ParticipantInfo be a pure component (only takes props) and then when it needs to be reused by other things like visit, medical data etc, those should be parent components passing props to it.

crsmyz:
@MikeyManoguerra In regards to your comment about adding the future functionality to ParticipantInfo, I would like to see that most if not all of that logic handled by the ParticipantInfo parent itself, and then the functional views can handle the rest. Similar to how the ParticpantList is handled. This approach should make unit testing a lot simpler as well, concentrating all complexity and data flow into a few nodes as opposed to the reusable child components.

MikeyManoguerra:
closed to wait for a more focused form refactor

PR #302, opened 2020-05-12
added indexes for search

mariekers:
I did it!  I coded!


MikeyManoguerra:
I believe you have to run `python manage.py makemigrations`for this to take effect. You'll see a new migration file afterwards

mariekers:
added migration - also spun up the back end to make sure it still worked

PR #304, opened 2020-05-14
fixes bug created by a previous PR of mine. Move adding participant id...

MikeyManoguerra:
to visit POST, align with checking for exiting visit when component is loaded.
This allows the current implementation to recognize there is no visit (coming from ParticipantList) and then create on successfully. also fix placeholder Notes field


MikeyManoguerra:
Taking this down, make more complete PR later

PR #308, opened 2020-05-18
PR298: Changing refs from user to participant

crsmyz:


crsmyz:
Yea I just did a search throughout the files in VS code so its unlikely I missed any. Also went through each file individually as well.

PR #309, opened 2020-05-19
PR297: Toggle search bar on submit + optimizations

crsmyz:


frontend/src/components/ParticipantSearch.js
```diff
@@ -35,7 +34,7 @@ const ParticipantSearch = observer(() => {
       <Grid
         container
         component="form"
-        onSubmit={handleSubmit}
+        onSubmit={e => (handleSubmit(e), props.handleToggle())}
```
coneill-relay, line 37: This seems odd to use comma syntax like this.  `handleToggle` should just be part of the handleSubmit code

ThrowsException, line 37: Sorry this is the thrid time now. lets see if I can get this right. the `handleToggle` should just be moved into the handleSubmit function.

Issue #312, opened 2020-05-29
WIP(ticket):schema refactor


MikeyManoguerra:
**Ticket is work in progress**

in no particular order or weight
- unspool and remove program-service-map
- create a <program>_data for each program which will have the visit id and either:
        A. roughly represent a corresponding form
        B.  be a link tree of foreign keys to point to subdata tables ( maybe only this to future proof??)
- permission data and "subdata" tables accordingly

MikeyManoguerra:
program service map remove, see forthcoming tickets for _data work

Issue #279, opened 2020-04-29
add output to seed.py


mariekers:
`seed.py` has all output notications turned off, and has the appearance that nothing is happening.
- [ ] add a print statement that says : 'starting seed' or something similar
- [ ] add a print statement that says : 'seed completed' or something similar
- [ ] optionally decide to turn on or add more descriptive output statements that are already built in to the seed functions

PR #317, opened 2020-06-19
Unspool service_program relation in preparation for program specific program_data  tables

MikeyManoguerra:
- remove program service map 
- service has program fk
- update visit save() to check for valid program service pair
- update tests to reflect changes


MikeyManoguerra:
#279 and some of #312

Issue #321, opened 2020-06-23
Reusable Handle change method


mariekers:
Make "handle change" onChange method for form inputs reusable
Currently, Each form field has its own individual method. 

Update so there is a universal method that takes `e.Target` and uses `name` property and `value` property from input to update state.

See react docs:
https://reactjs.org/docs/forms.html#handling-multiple-inputs

LizKalter:
Hello. I'm interested in working on this issue.

MikeyManoguerra:
Awesome!

A couple of extra notes... this 'handleChange' method should replace the update type methods in the `ParticipantStore.js`, and thus you can pass less props down to the Form components like `<ParticipantForm/>`

reach out if you have any questions on what the app is doing  or how we handled something.

LizKalter:
Looks like some tests are failing because form field name attributes changed. Should I update the tests as part of the pull request?

PR #338, opened 2020-07-25
Issue #321

LizKalter:
Proposed changes for issue #321


frontend/src/stores/ParticipantStore.js
```diff
@@ -108,6 +117,18 @@ export class ParticipantStore {
     this.routeToQueueTable = data
   }
   // Participant State Actions
+  @action handleParticipantChange = event => {
```
MikeyManoguerra, line 120: Can you pass `e.target` to handleParticipantChange where it is called? 

then in handleParticipantChange, you can de structure  `const { name, value } = target`

and assign as such.
`this.participant[name] = value` 

unless there is a reason to do otherwise, there will be less repetition

frontend/src/components/ParticipantForm.js
```diff
@@ -126,25 +116,23 @@ function ParticipantForm(props) {
...
       </Grid>
       <Grid item xs={12} sm={6}>
         <FormControl>
           <InputLabel id="insurance-select">Select Insurance</InputLabel>
           <Select
-            name="insurance"
+            name="insurer"
             value={participantInfo.insurer}
-            onChange={e => setInsurer(e.target.value)}
+            onChange={e => handleParticipantChange(e)}
```
MikeyManoguerra, line 135: pass  `e.target`  here, see explanation below

frontend/src/stores/ParticipantStore.js
```diff
@@ -108,6 +117,18 @@ export class ParticipantStore {
...
+    switch (event.target.name) {
```
MikeyManoguerra, line 121: Switch is ok here because these forms will not be changing that much over time. We will have to figure out a different way to do transformations on the 'data' forms (if at all). This is a thinking out loud comment, not a request for change

LizKalter: One issue here is that radio buttons return a string value and we need a boolean.

MikeyManoguerra: Yea totally. I thought about asking for a custom method in ParticipantInfo that handles this and keeping the handleChange without logic, but that's not really scalable for dynamic forms either.  

The "data" forms are purposely going to rely heavily on Selects for data integrity, and we can handle cascading effects in useEffect hooks (or similar) instead of the on change method.  Again, just thinking out loud

PR #339, opened 2020-07-26
Fix notes dialog

### LINKED ISSUE #334: Notes in notes dialouge on queue not displaying notes
The notes are not displaying correctly in the notes dialogue. Investigate the QueueStore logic.

 It might be a quick fix on the filtering logic.

ThrowsException:
fixes #334 
Also cleaned up the package.json a little bit and removed some unneeded deps


ThrowsException:
@MikeyManoguerra yea just the `observer` seemed to be the issue. 

I noticed the issue with the batching only on the PR tests actually. Locally they ran fine without it but until I put that in, the tests for the snapshots kept failing. Their docs seem to suggest that it should be on all the time so I didn't see any adverse affects to adding it. 

I'm only on the package.json tear because I just cleaned up a ton of other projects that had everything in dev deps 😬  ha

MikeyManoguerra:
let the good times roll!

PR #341, opened 2020-07-29
clean up babelrc

ThrowsException:


ThrowsException:
Yea it looks like Babel preset brings it in. I hate relying on other projects dependencies but it's probably fine to axe ours in this case. I've been looking to get back to this pr

MikeyManoguerra:
ok once you remove it from the .babelrc ill approve it

ThrowsException:
This should be good I pulled it out of package.json. relying on babel version now

PR #342, opened 2020-08-01
Header refactor

LizKalter:
Proposed fixes for issues with multiple headers and nested headers.


frontend/.babelrc
```diff
@@ -3,6 +3,6 @@
   "plugins": [
     ["@babel/plugin-transform-runtime", { "regenerator": true }],
     ["@babel/plugin-proposal-decorators", { "legacy": true }],
-    ["@babel/plugin-proposal-class-properties", { "loose": true }]
+    ["@babel/plugin-proposal-class-properties", { "loose": true }, "plugin-proposal-class-properties-alias"]
```
LizKalter, line 6: I'm not sure if we want this change merged, but I needed it in my environment for the frontend to build successfully.

frontend/src/components/Navbar.js
```diff
@@ -12,12 +13,58 @@ import MenuItem from "@material-ui/core/MenuItem"
 import AccountCircle from "@material-ui/icons/AccountCircle"
 import NotificationsIcon from "@material-ui/icons/Notifications"
 import Typography from "@material-ui/core/Typography"
+import MenuIcon from "@material-ui/icons/Menu"
 import GsTitle from "./GsTitle"
+import clsx from "clsx"
+
+const NavHeader = observer(({ drawerOpen, handleDrawerOpen, drawerWidth }) => {
```
LizKalter, line 20: I moved the drawer into the Navbar component to fix some of the header nesting issues. Needed to pass some props down for this.

frontend/src/components/Navbar.js
```diff
@@ -12,12 +13,58 @@ import MenuItem from "@material-ui/core/MenuItem"
...
+  const useStyles = makeStyles(theme => ({
+    appBar: {
+      transition: theme.transitions.create(["margin", "width"], {
+        easing: theme.transitions.easing.sharp,
+        duration: theme.transitions.duration.leavingScreen,
+      }),
+      "& .navbar__logo": {
```
LizKalter, line 27: I moved some of the css from the external style sheet into the component.

frontend/src/components/Navbar.js
```diff
@@ -48,6 +95,20 @@ const NavHeader = observer(() => {
...
+        aria-label="open drawer"
+        onClick={handleDrawerOpen}
+        edge="start"
+        className={clsx(classes.menuButton, drawerOpen && classes.hide)}
+      >
+        <MenuIcon />
+      </IconButton>
+    )
+  )
+
```
LizKalter, line 111: The hamburger icon.

frontend/src/layouts/MainLayout.js
```diff
@@ -72,30 +48,19 @@ const useStyles = makeStyles(theme => ({
     }),
     marginLeft: 0,
   },
+  navbarOffset: {
+    ...theme.mixins.toolbar,
+    "@media (max-width: 700px)": {
+      minHeight: "116px",
+    },
+  },
```
LizKalter, line 56: I added an offset element so that the fixed header doesn't overlap the top of the page.

frontend/src/layouts/MainLayout.js
```diff
@@ -72,30 +48,19 @@ const useStyles = makeStyles(theme => ({
...
-          edge="start"
-          className={clsx(classes.menuButton, open && classes.hide)}
-        >
-          <MenuIcon />
-        </IconButton>
-        <Navbar />
-      </Toolbar>
-    )
-  )
-
```
LizKalter, line 98: I moved the hamburger icon into the Navbar component.

frontend/src/routes/index.js
```diff
@@ -12,28 +11,31 @@ import NewParticipantView from "../views/NewParticipantView"
 const Routes = () => {
   return (
     <BrowserRouter>
-      <Navbar />
```
LizKalter, line 15: I removed the Navbar here because it was a duplicate header on the page.

frontend/src/routes/index.js
```diff
@@ -12,28 +11,31 @@ import NewParticipantView from "../views/NewParticipantView"
...
-      <Route path="/login" component={LoginForm} />
-      <MainLayout>
```
LizKalter, line 17: I also added an additional Switch and Router around MainLayout so that the MainLayout component isn't on the page at the same time as the LoginForm Component.

frontend/src/style/main.scss
```diff
@@ -1,5 +1,3 @@
 // import component scss files here. main is imported in App.js
 
-@import 'login-form';
-@import 'navbar';
```
LizKalter, line 4: I moved the css from these files into the components.

frontend/src/views/LoginForm.js
```diff
@@ -25,49 +63,67 @@ const LoginForm = observer(({ location }) => {
...
-          )}
+        <Toolbar className={classes.navbarToolbar}>
+          <img
+            src={PreventionPointLogo}
+            alt="Prevention Point Logo"
+            className="navbar__logo"
+            width="150"
+          />
+        </Toolbar>
+      </AppBar>
```
LizKalter, line 80: The diff looks a little weird, but all I did was add a basic header with just the logo in it. It didn't look like any of the other header elements were functional on the login page, so I didn't think they were necessary. I tried just including the Navbar component, and while the site seemed to function fine this way, I got a test error back saying that withRouter was being used outside of a Route.

frontend/src/components/Navbar.js
MikeyManoguerra: This is good. We are using so little SCSS I would like to remove it if feasible.

frontend/src/routes/index.js
MikeyManoguerra: I think you executed your approach well. We were going to go in a slightly different direction, though it has stalled as of late [this PR](https://github.com/CodeForPhilly/prevention-point/pull/309) was the plan, though it has not been finished. 

If you were to remove the Main layout entirely, and make the auth routes different only by having this containing drawer component (that would live in /components/), and finally, left the nav bar as is (and persistent at the top of the router, would that be easier, or still work?

The reason I bring this up is for the foreseeable scope, we will only have one layout. removing this pattern will streamline the directory structure a little.

frontend/.babelrc
MikeyManoguerra: It would be better to merge the fix PR before we merge this PR.

frontend/src/routes/index.js
LizKalter: I'm not sure that I completely understand your suggestion yet. Are you thinking of essentially renaming MainLayout to Drawer and then wrapping both the login and the private route components in it? I am not sure if I will encounter the same error I had when I put the Navbar component into the Login page. It seemed to take issue with having the withRouter call that determines whether to display the hamburger. I'm not sure why though since there was already another withRouter call on the page to begin with. Anyway, I can certainly give it a try.

LizKalter: I just made a new commit with MainLayout as the container for all of the routes including the login page. The Navbar is included inside MainLayout as it is dependent on some of the internal state from MainLayout for displaying the hamburger icon. I am not seeing any errors with this arrangement. Am I going in the right direction?

MikeyManoguerra: After looking a little closer at this group of components, its a little more convoluted than I thought. 
I Think you were thinking about this in the right way, (wrapping with the layout), but I'm going to formulate a slightly different refactor approach. 

Ill write it out in more detail tomorrow, but the gist of it is:
- navbar conditionally renders based upon whether the user is logged in with `rootstore.isauthenticated`, not history. 
- if still needed, any places that have `withRouter` should be refactored to `useRouter`
- no nested switch.
- change privateRoute.js so  <Component/> is wrapped in the Drawer component
- remove navbar from drawer.

sorry this is a clipped response, but I just want to jot this down, and stop you from doing too much more without direction. ill send you a larger response tomorrow

MikeyManoguerra:
OK, So after closer inspection, I'll have you update this in two parts. first this router stuff, and then go back with a second ticket and do some component refactoring.In the end, it will be flatter, and easier to style.

Ignore statements about layout in previous comments. We will keep the 'Layout' pattern we are currently using. I think the below description will be easiest to implement and work with, (and we can keep the props you made being passed down from layout into navbar)

##### 1. Add the layout to PrivateRoute.js
- import `<Mainlayout/>` into PrivateRoute.js.  wrap  the `<Component/>` as so: 
``` 
...
 rootStore.authStore.isAuthenticated ? (
          <MainLayout>
            <Component {...props} />
          </MainLayout>
        ) : (
...
```
- here are [the docs](https://reactrouter.com/web/api/Route/render-func)
- add a comment in PrivateRoute.js that says: `// to use programmatic layouts, give this custom route component a 'layout' prop, and pass the layout like we are passing 'component'`

##### 2. Public Route
- create a new file called PublicRoute.js .
- this should be very similar to the private route, (with the Main Layout wrapping the component). The only difference is does not check for authentication, and has no redirect. 

##### 3. flatten browser router
- the layout and nav are handled in our custom routes, so you can flatten the BrowserRouter.
```
 <BrowserRouter>
      <Switch>
        <PublicRoute />
        <PrivateRoute />
       ...
      </Switch>
   </BrowserRouter>
```

##### 4. 
Now, you can conditionally render hamburger. I recommend switching the logic to render based upon `rootStore.authStore.isAuthenticated`

you should look at how the other components access history with `useHistory` (participantInfo.js). This is  the more modern way to access router state.


Like I said at the top, This will get the layout and navbar issues in order, then I will make another ticket with some refactor thoughts I have for the actual JSX in the components.

If anything does not make sense, let me know. I'm around today.

LizKalter:
@MikeyManoguerra Thank you for the detailed feedback. I think that my latest commit should cover everything discussed.

Issue #336, opened 2020-07-24
Live Search on Participant List


MikeyManoguerra:
On The ParticipantList page, we want to add the functionality for onChange live search in a search bar. As a user starts typing a name, the participant list will filter the results.

This will be used for quick form fill outs. If a participant is waiting in line for a short service ( syringe exchange, receiving mail) 
A user can:
- filter the users
- take them to the more robust visits page
- OR, use a dropdown(?) of "quick forms" that can have a modal popup (to be determined)

This requires a bit of a redesign of the participantsList.
- keeping a persistent list of all participants in state while the app is in use
- lazy loading the rows. React will be overloaded if it has to re-render thousands of participants on every keypress

If you have ideas for this issue, please share

LizKalter:
I'm looking into an implementation using Material UI Autocomplete. It may not meet all the requirements here, but I think it's a start.

mariekers:
check to see if we should be avoiding the model serializer?

MikeyManoguerra:
As discussed last week, this can delay an api call until the user enters the first few query characters. This can then use the endpoint to get a subset of participants, and continue to filter with js. if the query string gets below our threshold length, it will re request data from the api once the minimum query length is reached again

PR #343, opened 2020-08-02
Participant autocomplete

LizKalter:
Proposed changes for issue #336, or at least a starting point for it. This change introduces an out-of-the-box Material UI Autocomplete component for filtering the participant list.


frontend/src/views/ParticipantList.js
```diff
@@ -65,9 +81,55 @@ const ParticipantList = observer(() => {
...
+              <Grid item>
+                <SearchIcon className={classes.searchIcon} />
+              </Grid>
+              <Grid item>
+                <Autocomplete
+                  disableClearable
+                  noOptionsText="No results found"
+                  options={participantStore.participants}
+                  getOptionLabel={participant => {
+                    return participant.first_name + " " + participant.last_name
```
ThrowsException, line 109: prefer string templates here `${participant.first_name} {participant.last_name}`

frontend/src/views/ParticipantList.js
```diff
@@ -65,9 +81,55 @@ const ParticipantList = observer(() => {
...
+        className={classes.autocompleteWrapper}
+      >
+        <Grid item>
+          <PrevPointHeading className={classes.participantsListHeading}>
+            Participants
+          </PrevPointHeading>
+        </Grid>
+
+        <Grid item>
+          {participantStore.participants.length > 0 && (
```
ThrowsException, line 98: just `participantStore.participants` is good enough here. if the value is truthy it will display.

Issue #330, opened 2020-07-24
Form, Question, and Option Database Tables


MikeyManoguerra:
Use the Django Model Classes to make the Database tables for dynamic forms
Each model should get its own file in core/forms

https://docs.google.com/spreadsheets/d/1-iD4m4jzEAE8EEeEY_hPXYiplpKz74hrPlI5TZY7bVM/edit#gid=0
The tabs at the bottom relate to the tables. reach out if you need clarification on data types

PR #344, opened 2020-08-04
Form, Question, and Option Database Tables 

pldimarco:
PR for: https://github.com/CodeForPhilly/prevention-point/issues/330


core/models/option.py
```diff
@@ -0,0 +1,9 @@
+from django.db import models
+from . import Program, Form, Question
+
+class Option(models.Model):
+    program = models.ForeignKey(Program, null=True, on_delete=models.CASCADE)
```
MikeyManoguerra, line 5: having program is somewhat redundant here, will confirm

core/models/question.py
```diff
@@ -0,0 +1,21 @@
+from django.db import models
+from enum import Enum
+from . import Program, Form
+
+class Type(Enum):
```
MikeyManoguerra, line 5: I want to be explicit when the question should render as 'select'

- there should be another field "input_type" that takes an enum choice, which is the type of element the dom should render.
- if it is 'select' (our most common field) , then we will include the options for that question in the response. 

other input_types include: radio, text, textarea, and date. we dont need an exhaustive list, these most common ones will do for now

core/models/option.py
pldimarco: Yeah, I did see that all the information was on every table. But rather than assuming we would be scanning the table for the key, I included the foreign keys on the table like in the google doc.

core/models/question.py
pldimarco: Yeah. I tried to see if I could pull the table types out of somewhere to be used instead of the enum, but I didn't find anything easily accessible. 

But for the `input_type` part of your comment, do you mean the current `Enum` should just have a `select` option that would direct the UI to query the `options` table?

core/models/option.py
MikeyManoguerra: We can remove program id from `option` and `question`. having program id on `form` is sufficient, and then , we filter by form id. i see this working as you query form table to get a list of forms for that program, and then use the selected form's id to go get the actual form questions.

core/models/question.py
MikeyManoguerra: When i wrote that I was thinking two enums, but I think that is overkill. I know that the sheet means data type, But I think the more valuable thing to store on this table might be the input type the front end needs. 
- select
- text
- date
- textarea
- number
- radio (for booleans)

That way, its explicit for the frontend and it doesnt have to make inferences, just render.
I dont think the back end needs such specific data type definitions. It going to be text, a number, or a boolean. 

Does this make sense?

pldimarco: I believe so. So we're not trying to have the column hold a specific SQL data type, but rather some generic data types for the input type. And we just want an `Enum` of the list in your previous comment?

MikeyManoguerra: yes,  where the data is saved can handle the data type,  and we can also check with the future validators table.

pldimarco: Changes pushed

pldimarco:
@MikeyManoguerra Pushed the changes from your comments. Just pinging to get some eyes on this again.

Issue #347, opened 2020-08-07
Nav/sidebar component cleanup


MikeyManoguerra:
- Inline the  variable sub component buttons into return statement JSX [like this and similar](https://github.com/CodeForPhilly/prevention-point/blob/fc4c9029461e74a793fab37ca388ab79d53f3396/frontend/src/components/Navbar.js#L70)
- Do we need both carrots on the sidebar? ( if they both show up in the UI, keep them, if only one shows, remove the one that points open, as we are using the the hamburger to open the sidebar.)

- make sure the sidebar closes after clicking "search"

MikeyManoguerra:
@LizKalter here is a ticket of the issue I described, if you want to take it

Issue #335, opened 2020-07-24
Participant table needs margin bottom


MikeyManoguerra:
When you scroll to the bottom of the participant table, The last entry is cut off by the "add participant" button. 

A margin bottom on the participant table will account for this

PR #348, opened 2020-08-08
Feature/ui updates liz

LizKalter:
Updates for issues #347 and #335


frontend/src/components/Navbar.js
```diff
@@ -149,7 +114,32 @@ const NavHeader = observer(({ drawerOpen, handleDrawerOpen, drawerWidth }) => {
...
+          onClose={handleProfileMenuClose}
+        >
+          {rootStore.authStore.isAuthenticated ? (
+            <div>
+              <MenuItem onClick={handleProfileMenuClose}>Profile</MenuItem>
+              <MenuItem
+                onClick={() => {
+                  handleProfileMenuClose()
+                  rootStore.authStore.logout()
+                  history.push("/")
```
ThrowsException, line 131: nit: Prefer not to have inline functions doing multiple things id suggest moving the
 ```
rootStore.authStore.logout()
history.push("/");
```
into the `handleProfileMenuClose` so that its all encapsulated in one function

ThrowsException:
overall looks good. one minor change.

frontend/src/components/Navbar.js
LizKalter: That would cause the user to get logged out whenever the menu closes. Maybe you want a new function called logoutUserAndCloseMenu or something like that?

ThrowsException: youre right. actually for logout this is fine.

PR #362, opened 2020-08-22
Multiline notes field

### LINKED ISSUE #360: Use the multiline flag on notes input.
Update the VisitForm notes field to use PrevPointInput and multi line flag

https://material-ui.com/api/input/

https://github.com/CodeForPhilly/prevention-point/blob/e558848c7b763f307521ccc0b2da16fb2c8e40ca/frontend/src/components/VisitRouter/VisitForm.js#L84

LizKalter:
Fixes #360


frontend/src/components/VisitRouter/VisitForm.js
```diff
@@ -81,13 +81,17 @@ const VisitForm = ({
...
-          label="Add a Note"
-          name="notes"
-          value={visitInfo.notes}
-          onChange={e => handleVisitChange(e.target)}
-        />
+        <FormControl>
+          <InputLabel htmlFor="addANote">Add a Note</InputLabel>
+          <PrevPointInput
+            fullWidth
+            multiline="true"
```
MikeyManoguerra, line 88: I think you can just put 'multiline' by itself(not equals true). like the fullwidth flag above.

MikeyManoguerra, line 88: you can just put multiline as a flag, i believe, you do not need to add `=true`

LizKalter: Sure enough.

PR #363, opened 2020-08-30
add unique sep_id field to participants, write seed and tests to accompany it

MikeyManoguerra:


core/tests/participants.py
```diff
@@ -1,28 +1,9 @@
-from django.contrib.auth.models import Group, User
+import json
 from core.tests.base import BaseTestCase
-from core.models import Participant
-from core.models.insurer import Insurer
+from core.models import Participant, Race, Gender
 
 class ParticipantsTestCase(BaseTestCase):
-    def setUp(self):
```
mariekers, line 7: does this remove 'insurer'?  Cause I'm gonna be honest - it's pretty unnecessary.

MikeyManoguerra: just in the test

Issue #365, opened 2020-09-02
Change this method name to createNewFrontDeskEvent


MikeyManoguerra:
https://github.com/CodeForPhilly/prevention-point/blob/a071e461e12d305931463241352c8969af91ff34/frontend/src/stores/ParticipantStore.js#L256

update all places it is used

JackRyan1989:
Merged.

PR #369, opened 2020-09-08
Renamed getFrontDeskEvents method to createNewFrontDeskEvents in 

JackRyan1989:
frontend.

### DIFF
--- a/frontend/src/api/frontDeskEventEndpoints.js
+++ b/frontend/src/api/frontDeskEventEndpoints.js
@@ -1,4 +1,4 @@
-export const createNewFrontDeskEvent = api => async id =>
+export const getFrontDeskEvent = api => async id =>
   await api.get(`/front-desk-events/${id}/`)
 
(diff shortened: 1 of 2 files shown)


ThrowsException:
why `createNewFrontDeskEvent` when the operation is a get? That seems like it would imply a PUT or POST operation and make the `postFrontDeskEvent` confusing.

ThrowsException:
greping the code it doesn't even look like `getFrontDeskEvent` is even used can it just be removed?

JackRyan1989:
This is in response to Issue #365, I'm not sure what the logic was initially behind the request to change the name.

JackRyan1989:
If there's no plan to use the command in the future I'm ok with removing the method.

JackRyan1989:
@ThrowsException Ok, I changed back the api endpoints, and just made a small change in the ParticipantStore.js file to make the error messages a consistent format. Let me know if you have any questions or concerns.

ThrowsException (approved):

Issue #370, opened 2020-09-11
Update styling on "disabled" inputs


MikeyManoguerra:
Use the css ":disabled" selector to update the text color of disabled elements. 

The place where they are currently being used is the ParticipantForm.

The current color is a little too light to be read easily, and the assumption is they will not be updating these fields very often, but will be seeing them every time they are working with an individual's data.

MikeyManoguerra:
@LizKalter Heres the updated ticket, i hope you are feeling better!

PR #371, opened 2020-09-12
Update disabled text color for issue #370

LizKalter:
Stylesheet changes for issue #370. (Please ignore the branch name here. It is the old ticket number.)


MikeyManoguerra:
Hey! Wondering if you can update the disabled coloring globally here instead: https://github.com/CodeForPhilly/prevention-point/blob/master/frontend/src/style/theme.js

LizKalter:
I tried that first and the color applies to everything that gets the Mui-disabled class. I didn't think it looked good for the labels to change color as well.

MikeyManoguerra (approved):

PR #373, opened 2020-09-15
Added this.setRouteToQueue(false) to createNewFrontEndDeskEvents, cre…

JackRyan1989:
…ateParticipant and updateVisit methods in ParticipantStore.

@MikeyManoguerra I decided to add a small if check when setting the flag to false, I'm actually not certain it's needed at all since we always want to set the flag to false initially, just thinking maybe it's best to not alter state if it's not needed? Let me know if I'm not making sense or am being overly cautious or something.


MikeyManoguerra (approved):
I'm ok with the conditional check

PR #382, opened 2020-09-23
add SEP data table, remove "admin" group in favor of is_superuser

MikeyManoguerra:
This creates the SEP data model. 
I also removed the  "Admin" group and permissions. This was redundant because the admin is already a superuser, so there is no need to maintain a list of admin privileges. the admin has all privileges by default.


ThrowsException (approved):

JackRyan1989 (dismissed):

MikeyManoguerra:
dang need a re-review, rebase not smooth.

PR #390, opened 2020-09-28
Site route created on Backend

JackRyan1989:
Files created:
1. /core/fixtures/sites.yaml
2. /core/sites/serializer.py
3. /core/sites/views.py
4. /core/tests/site.py
5. /core/migrations/0005_auto_20200927_2021.py

Files changed:
1. /management/command/seed.py
2. /management/command/users_and_groups.py
3. /management/command/seed.py
4. /management/command/users_and_groups.py
5. /core/models/sep_data.py 
6. /core/urls.py
7. /core/tests/__init__.py


core/management/commands/users_and_groups.py
```diff
@@ -44,6 +44,7 @@
     "add_behavioralhealthnotes",
     "change_behavioralhealthnotes",
     "view_behavioralhealthnotes",
+    "view_sites",
```
MikeyManoguerra, line 47: This should be 'site' singular, to match the new model name.

JackRyan1989: Gotcha, so these are only ever going to be displaying a single site? I was trying to make it match whether you could be viewing one or more sites, but I'll just make it match the model.

MikeyManoguerra: Permissions are not that fine grained out the box. There are four available ones, which are verb/CRUD based. 
https://docs.djangoproject.com/en/3.1/topics/auth/default/#default-permissions

Django REST Frameworks does not permission GET requests, like MVC Django does, so in `core/permissions.py`, we modified the class slightly to enable this feature.

core/models/site.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+from django.utils import timezone
+
+class Site(models.Model):
+    site_name = models.CharField(max_length=100)
+    site_type = models.CharField(max_length=100)
+    description = models.CharField(max_length=100)
+    address = models.CharField(max_length=1000)
+    zip_code = models.CharField(max_length=10)
+    created_date = models.DateTimeField(default=timezone.now)
```
MikeyManoguerra, line 10: I should have caught is before, can you make this `created_at`? as to match similar fields across the app.

core/tests/site.py
```diff
@@ -0,0 +1,66 @@
...
+
+        response = self.client.get('/api/site/{}'.format(site_id), follow=True, **headers)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        self.assertContains(response, 'TESTING1')
+        self.assertEqual(7, len(response.data))
+        self.assertNotContains(response, 'TESTING2')
+        self.assertNotContains(response, 'alskdjfuuru')
+
+    # def test_admin_filter_site_by_params(self):
```
MikeyManoguerra, line 49: remove this if not testing

MikeyManoguerra (changes requested):
couple of small things, looks good over all. when updating the model field, you can delete the 0005 migration file and generate a new one

core/tests/site.py
JackRyan1989: This test currently isn't working as I expect, but I left it in case we want to hit this route and use params? If that isn't likely I can delete.

core/models/site.py
JackRyan1989: This change is reflected in the new commit.

JackRyan1989 (commented):
Changes made and present in latest commit.

MikeyManoguerra (approved):

PR #399, opened 2020-10-05
Responsive design

dmjohnspor:


MikeyManoguerra:
A few things. 
- the front end test suite failed. i looked through the logs, and it comes from the aria label we changed. update the test to match the arialabel (`thead` to `table head`).
- this is a good time to mention its best to run the tests before you commit. we will be working on tests together soon
- You based this branch off of the `error-handling` branch, and thus all that work is also present here. That's ok, lets get the other branch merged in first though. we will talk about git this week, to align on some conventions!
- This one is completely on me, but i would like to use `title` as the key, instead of `name`, in the constants object after seeing the updated component. so update the constant `PARTICIPANT_LIST_TABLE_TITLES` key to titles, de structure `title` from the map fn value,and so on.

MikeyManoguerra:
once the `error-handling branch` is merged,

`git fetch` to get the updated master
`git rebase origin/master` to update your branch with the new work

MikeyManoguerra (approved):

Issue #384, opened 2020-09-23
SEP data route


MikeyManoguerra:
depends upon #382 

- Create a new directory `sep_data` that has a view and serializer.  do not worry about populating the data for the FKs, this will primarily be a POST endpoint at first
- register the route in `urls.py`, under default router

##### In `management/commands/users_and_groups.py`  
- give add  permission to front desk
- give add, view, and change permission to internal provider

then 
- write a couple of tests . you will have to create fixtures. 
- make sure there is a POST test

MikeyManoguerra:
@sxgormley  this is a good example, if  you see this merged flag, its probably safe to close the issue.  There might be some times that a PR does not complete everything in an issue, so watch out for that. However, most of the time 1 issue === 1 PR, like here

sxgormley:
How do I tell if a PR doesn't complete everything in an issue?

sxgormley:
> @sxgormley this is a good example, if you see this merged flag, its probably safe to close the issue. There might be some times that a PR does not complete everything in an issue, so watch out for that. However, most of the time 1 issue === 1 PR, like here

How do I tell if a PR doesn't complete everything in an issue?

MikeyManoguerra:
ideally people will tag and mention that it does. for the scope of the tickets over the next month, its pretty much safe to assume it does unless explicitly stated otherwise. 

In addition my personal opinion is that the larger half finished ticket should be closed anyway and anew smaller ticket with the rest should be opened.  But yea, feel free to close unless it says don't !

PR #401, opened 2020-10-07
SEP Data route

MikeyManoguerra:
creates an endpoint to post SEP data to.
#384

### DIFF
--- a/core/queue/serializers.py
+++ b/core/queue/serializers.py
@@ -1,9 +1,0 @@
-from rest_framework import serializers
-from django.utils import timezone
-from core.models import Program
-
-
-class QueueSerializer(serializers.ModelSerializer):
-    class Meta:
-        model = Program
-        fields = ("id", "name", "services")
--- a/core/tests/sep_data.py
+++ b/core/tests/sep_data.py
@@ -1,0 +1,72 @@
+import json
+from django.utils import timezone
+from django.urls import reverse
+from rest_framework import status
+from core.models import Sep_Data
+from core.tests.base import BaseTestCase
+
+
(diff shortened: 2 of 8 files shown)


core/tests/sep_data.py
```diff
@@ -0,0 +1,72 @@
...
+            "needles_in": 5,
+            "needles_out": 5,
+        }
+
+        response = self.client.post(url, data, format="json", follow=True, **headers)
+        content = json.loads(response.content)
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertTrue(Sep_Data.objects.filter(pk=content["id"]).exists())
+
```
JackRyan1989, line 72: Not sure this is necessary, but maybe add a test to show that we cannot post using uds_provider? That would just be a complete test of the permissions set up.

core/queue/serializers.py
```diff
@@ -1,9 +0,0 @@
-from rest_framework import serializers
```
JackRyan1989, line 1: Why was this file deleted?

JackRyan1989 (commented):
Looks really good! Just had a few simple questions.

core/queue/serializers.py
MikeyManoguerra: this serializer was never used.  The queue just uses the already existing Program serializer , since there is no queue table.

core/tests/sep_data.py
MikeyManoguerra: I added one!

JackRyan1989: Perfecto!

JackRyan1989 (approved):

PR #410, opened 2020-10-10
Updated participant views to return 404 when queryset does not exist.

JackRyan1989:


MikeyManoguerra (approved):

MikeyManoguerra (changes requested):
can you write a test that makes sure it returns 404 with a search query? I think `retrieve` does return 404, but a `list` always return 200 by default

JackRyan1989:
@MikeyManoguerra I'm running into this error in trying the query test: AttributeError: 'Response' object has no attribute 'model'

I think that is due to the response I'm returning in the /participants/views.py, line 39. It's not in the same format as the querysets that are returned. 

Also, I'm not sure what you mean by retrieve?

MikeyManoguerra:
list  refers to and endpoint `/api/participants/`, which returns all or a subset of objects
retrieve then would be `/api/participants/<id>` where it returns a single object
https://www.django-rest-framework.org/api-guide/viewsets/#viewset-actions
---

This endpoint is overriding the `get_queryset` method, and not the `list` method. The ModelViewSet is actually passing this queryset to the uncustomized list method. Basically trying to send a response from this method is too early in the stack.

To be more consistent across the app we could rewrite as custom list(), but easier to throw a drf exception

https://www.django-rest-framework.org/api-guide/exceptions/#notfound


https://stackoverflow.com/questions/49738795/how-to-return-a-404-when-drf-returns-empty-queryset-using-listapiview

If you have not heard me complain about how DRF is strict, this is the kind of stuff I am talking about. But really , its because  `ModelViewSet`  and  `ModelSerializer` are hiding a ton of internal logic, they are very high level classes

core/tests/participants.py
```diff
@@ -124,4 +128,23 @@ def test_can_query_maiden_name(self):
 
         self.assertEqual(200, response.status_code)
         self.assertTrue(len(content) == 1)
-        
\ No newline at end of file
+
+    def test_return_four_oh_four(self):
+        headers = self.auth_headers_for_user("front_desk")
+        known_participant_id = 100000
```
MikeyManoguerra, line 134: this is minor but the variable name should be something more like invalid_id or out_of_range_id, something to indicate that its not in the db.

core/tests/participants.py
```diff
@@ -124,4 +128,23 @@ def test_can_query_maiden_name(self):
...
+
+        response = self.client.get(
+            f"/api/participants/{known_participant_id}", follow=True, **headers
+        )
+        
+        self.assertEqual(404, response.status_code)
+
+    def test_return_four_oh_four_query(self):
+        headers = self.auth_headers_for_user("front_desk")
+        known_participant_id = 100000
```
MikeyManoguerra, line 144: same here

core/tests/participants.py
JackRyan1989: Yup no worries! Sorry for being a bit sloppy.

core/tests/participants.py
JackRyan1989: Will do

core/tests/participants.py
MikeyManoguerra: its all good. I feel sloppy on my code reviews, I didnt notice it till now.

JackRyan1989:
@MikeyManoguerra Raising an exception worked. I think I have more questions about some of the things that came up during this ticket, let me know if you're around to chat next week or so. If not, that's cool too!

JackRyan1989 (commented):
Simplified test and used Exception when queryset is not found.

JackRyan1989 (commented):
Outdated but addressed.

JackRyan1989 (commented):
Addressed

JackRyan1989 (commented):
Addressed

JackRyan1989 (commented):
Addressed

MikeyManoguerra:
> @MikeyManoguerra Raising an exception worked. I think I have more questions about some of the things that came up during this ticket, let me know if you're around to chat next week or so. If not, that's cool too!

Sweet, I am around. I can try to answer!

MikeyManoguerra (approved):

Issue #407, opened 2020-10-09
Add New SEP ID and mothers maiden name field


sxgormley:
We added a couple of new fields to the Participant model, so now we have to update the front end to reflect this.
**NOTE** Im going to try and get all instances of this, but if i miss one and you spot it, it was probably a mistake on my part.
​
- below UUID and above race, add two new input fields with name atributes of `sep_id` and `maiden_name`
- add empty strings for these fields in `setDefaultParticipant` in the `participantStore`
- add a new header to the HEADER TITLES constant and the Participant Table body, the 'SEP id'.  do not add maiden name to the participant table at this time
- in `QueueTableComponent/QueueTable.js`, add maiden_name and sep_id to the `setParticipant` fn that starts on line 102

sxgormley:
Hey, @dmjohnspor ! How's it coming?

sxgormley:
Hey @MikeyManoguerra ! Does the Merged tag mean I can close this issue? Thank you!

PR #411, opened 2020-10-11
New participant input fields

dmjohnspor:


frontend/src/components/ParticipantTableComponent/PrevPointTableBody.js
```diff
@@ -47,6 +47,9 @@ const PrevPointTableBody = props => {
             <TableCell aria-label="tcell" className={classes.mobileVisibility}>
               <PrevPointCopy>{participant.date_of_birth}</PrevPointCopy>
             </TableCell>
+            <TableCell aria-label="tcell">
```
MikeyManoguerra, line 50: Can you add the mobile visibility class here? and also move it to left in the table, lets put it after last name

frontend/src/components/ParticipantForm.js
```diff
@@ -96,6 +96,32 @@ const ParticipantForm = ({
           />
         </FormControl>
       </Grid>
+      <Grid item xs={12} sm={6}>
```
MikeyManoguerra, line 99: This was not in the ticket, but while reviewing this, the inputs get minuscule before switching to the mobile view when the sidebar is open.

can you remove the `sm={6}` on each input that has it , and replace it with `md={6}`?
can you also do the same to the `VisitForm.js`? Don't forget to update the snapshot!

frontend/src/components/ParticipantForm.js
```diff
@@ -96,6 +96,32 @@ const ParticipantForm = ({
...
+        <FormControl disabled={isDisabled}>
+          <InputLabel htmlFor="sep_id">SEP id</InputLabel>
```
MikeyManoguerra, line 101: capitalize to be SEP ID

MikeyManoguerra (changes requested):
Looks good, a couple of small requests to address.

frontend/src/components/ParticipantTableComponent/PrevPointTableBody.js
dmjohnspor: I addressed the issue.

frontend/src/components/ParticipantForm.js
dmjohnspor: I addressed the issue.

frontend/src/components/ParticipantForm.js
dmjohnspor: The change was made.

frontend/src/constants/index.js
```diff
@@ -19,6 +19,7 @@ export const PARTICIPANT_LIST_TABLE_TITLES = [
   { value: true, title: "Last Name" },
   { value: false, title: "Gender" },
   { value: false, title: "DOB" },
+  { value: false, title: "SEP id" },
```
MikeyManoguerra, line 22: capitalize ID

MikeyManoguerra (changes requested):
Looks good except one minor thing. 
However, I pulled the branch down, and realized the ticket was incomplete. 
can you add `maiden_name` and `sep_id` in the map fn here:

https://github.com/CodeForPhilly/prevention-point/blob/99b5fd3c61f5c6c8b64d1e87f2fab4e33a6e4bdf/frontend/src/components/QueueComponent/QueueTable/QueueTable.js#L74

they are on `x.participant`

The result should be, when you click on the actions button on the left side of the queue row, the data is then populated on the participant form, when taken to `/exisitingparticipant/` . <-- ill try and include a description of the 'results' of your feature, so you might catch stuff like this

_NOTE_
This is a  good time to mention that the way this is implemented is not the best, and leads to errors exactly like this. we are going to the store (the source of truth) and assigning new data directly to the row, then relying on the row data for some future action. 

A better approach would be to only get the data we need to render the row on the outset. When we have a further action that requires more data, use the participant _d and go back to the store for that data. We will fix this if we have time, I just want to point it out so you don't take everything in this repository as gospel

MikeyManoguerra:
#407

frontend/src/constants/index.js
dmjohnspor: Done

dmjohnspor:
Addressed all the issues

MikeyManoguerra (approved):

PR #416, opened 2020-10-18
fix console errors and lingering obsolete logic

MikeyManoguerra:
- upgrade material-table to fix errors
- heed warning and check mobx array length
- remove temporary checks in forms

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -50,7 +50,7 @@
     "babel-polyfill": "^6.26.0",
     "clsx": "^1.0.4",
     "formik": "^2.1.5",
-    "material-table": "^1.39.2",
+    "material-table": "^1.69.1",
     "mobx": "^5.9.4",
     "mobx-react-lite": "2.0.7",
     "moment": "^2.24.0",
--- a/frontend/src/components/VisitRouter/index.js
+++ b/frontend/src/components/VisitRouter/index.js
@@ -32,7 +32,7 @@ const VisitRouter = observer(() => {
     // kick off api call for program list from Mobx
     participantStore.getPrograms()
     // populate the global visit object if empty (for new visits)
-    if (!Object.keys(participantStore.visit).length) {
+    if (!participantStore.visit.id) {
       participantStore.setDefaultVisit()
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
--- a/frontend/src/stores/ParticipantStore.js
+++ b/frontend/src/stores/ParticipantStore.js
@@ -220,9 +220,7 @@ export class ParticipantStore {
       throw errorMessage
     }
   })
-  // called on  =>  ParticipantInfo.js
-  // andVisit temporary hack, this logic should be seperated out when creating VisitView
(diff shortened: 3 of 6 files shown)


frontend/package.json
```diff
@@ -50,7 +50,7 @@
     "babel-polyfill": "^6.26.0",
     "clsx": "^1.0.4",
     "formik": "^2.1.5",
-    "material-table": "^1.39.2",
+    "material-table": "^1.69.1",
```
MikeyManoguerra, line 53: fixes onCellEditable and similar warnings in console

frontend/src/components/VisitRouter/index.js
```diff
@@ -32,7 +32,7 @@ const VisitRouter = observer(() => {
     // kick off api call for program list from Mobx
     participantStore.getPrograms()
     // populate the global visit object if empty (for new visits)
-    if (!Object.keys(participantStore.visit).length) {
+    if (!participantStore.visit.id) {
```
MikeyManoguerra, line 35: checks if an existing database visit is trying to auto populate the form (coming from queue). this is more to the point

frontend/src/stores/ParticipantStore.js
```diff
@@ -213,9 +213,7 @@ export class ParticipantStore {
       throw `ParticipantStore:  getParticipants() Failed  =>  ${error}`
     }
   })
-  // called on  =>  ParticipantInfo.js
-  // andVisit temporary hack, this logic should be seperated out when creating VisitView
-  createParticipant = flow(function*(andVisit = true) {
+  createParticipant = flow(function*() {
```
MikeyManoguerra, line 216: we separated this logic with the NewParticpantView and ExistingParticipantView

frontend/src/views/AllQueues.js
```diff
@@ -60,7 +60,7 @@ const AllQueues = observer(() => {
   }, [queueStore])
 
   useEffect(() => {
-    if (queueStore.queues[tabValue]) {
+    if (queueStore.queues.length && queueStore.queues[tabValue]) {
```
MikeyManoguerra, line 63: This gets rid of the `mobx array at 0` console warning.

frontend/src/stores/ParticipantStore.js
JackRyan1989: @MikeyManoguerra Looks like the 'andVisit' hack was also creating a visit, I'm not seeing that function in either NewParticipantView or ExistingParticipantView, are we just not creating a visit simultaneously with a participant or is that function going to be called elsewhere?

MikeyManoguerra: This is a relic, related to visits that would end up in a queue. Originally, participant and visit was just one form, so this was an intermediary variable to handle splitting the two forms. now, when you create a participant in the new Participant view, it will route to the existing participant view and display the new participant there. the user then has the option to create a visit at that time

JackRyan1989 (approved):
This looks good. I have no concerns.

PR #417, opened 2020-10-18
Add SEP Search Forms

LizKalter:


LizKalter (commented):
@MikeyManoguerra This PR is in progress, but I am hoping you can take a look and let me know if everything is on the right track. I added some additional questions at the code level.

There are two separate forms for search and needles. The needles form is disabled until a single participant has been selected, then the SEP search form is disabled. (If you want to test this locally, I found last name "Huffman" will return a single result.)

You mentioned wanting a clear button on the form, but I'd like more direction on its placement and what it should do. I can see putting one next to the search button after a single participant has been selected in order to reenable to search form. Is this what you had in mind?

I have not added the submit functionality or validation to the needles form yet. What should happen after submitting the needles form? Should there be a success message? We could use the MUI snackbar if you just want a temporary message. https://material-ui.com/components/snackbars/#snackbar

frontend/src/components/SepForm.js
```diff
@@ -16,19 +26,203 @@ const useStyles = makeStyles({
...
+                      {[...Array(20)].map((x, i) => (
+                        <MenuItem key={i + 1} value={i + 1}>
+                          {i + 1}
+                        </MenuItem>
+                      ))}
+                    </Select>
+                  </FormControl>
+                </Grid>
+                <Grid item xs={6}>
+                  <FormControl disabled={participantId ? false : true}>
```
MikeyManoguerra, line 192: this can just be `disabled={!participantId}` instead of ternary

frontend/src/components/SepForm.js
```diff
@@ -16,19 +26,203 @@ const useStyles = makeStyles({
...
+          mothers_last_name: "",
+        }}
+        validationSchema={SEPSearchSchema}
+        onSubmit={async (values, { setSubmitting, setFieldValue }) => {
+          await participantStore.getParticipants({
+            //site_id: values.site_id,
+            sep_id: values.sep_id,
+            last_name: values.last_name,
+            date_of_birth: values.date_of_birth,
+            //mothers_last_name: values.mothers_last_name,
```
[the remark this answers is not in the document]
MikeyManoguerra: You should be able to search on 
- `sep_id`
- `last_name`
- `date_of_birth`
- `maiden_name` <- matches db model

frontend/src/components/SepForm.js
```diff
@@ -16,19 +26,203 @@ const useStyles = makeStyles({
...
-        <PrevPointHeading className={classes.heading}>
-          SEP Search
-        </PrevPointHeading>
-      </Grid>
+      <Formik
+        className={classes.form}
+        validateOnChange={false}
+        validateOnBlur={false}
+        initialValues={{
+          site_id: participantStore.getSite,
```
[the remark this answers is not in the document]
MikeyManoguerra: The endpoint is incorrectly named, it should be `/api/sites/` plural.  it returns an array.  (we can fix it across the full stack once this is done, just use `/api/site/` for now) 

- add a use effect that calls the site endpoint when mounted.  save results to `participantStore.sites`
- create a `participantStore.currentSite` variable. This is updated by  a select dropdown. The below input should become a select
```
  <FormControl className={classes.siteId}>
                  <InputLabel htmlFor="site_id">Site ID</InputLabel>
                  <PrevPointInput
                    id="site_id"
                    name="site_id"
                    readOnly
                    value={values.site_id}
                  />
                </FormControl>
```

- The selected site should remain the same throughout the session, hence saving it to global state. 
- If it causes problems being inside the Formik form, you can move it outside. I cant give a clear answer without playing around with this.

frontend/src/components/SepForm.js
```diff
@@ -16,19 +26,203 @@ const useStyles = makeStyles({
...
+                    name="mothers_last_name"
+                    onChange={handleChange}
+                    value={values.mothers_last_name}
+                  />
+                </FormControl>
+              </Grid>
+              <Grid item xs={12}>
+                <PrevPointButton
+                  type="submit"
+                  disabled={(participantId ? true : false) || isSubmitting}
```
MikeyManoguerra, line 165: this can just be disabled={participantId} instead of ternary

frontend/src/validation/index.js
```diff
@@ -23,4 +23,50 @@ const searchSchema = Yup.object().shape(
   ]
 )
 
-export { searchSchema }
+const SEPSearchErrorMessage = "Please enter a value in at least one field"
+const SEPSearchSchema = Yup.object().shape(
```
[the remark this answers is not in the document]
MikeyManoguerra: I do not know, that's more a question for @mariekers . This works for now, for me

MikeyManoguerra:
> There are two separate forms for search and needles. The needles form is disabled until a single participant has been selected, then the SEP search form is disabled. (If you want to test this locally, I found last name "Huffman" will return a single result.)

This works for me, less complicated then a multi-submit form. 

> You mentioned wanting a clear button on the form, but I'd like more direction on its placement and what it should do. I can see putting one next to the search button after a single participant has been selected in order to reenable to search form. Is this what you had in mind?

Yes that is the idea. Put it down at the bottom next to the submit button, though. Even though we know its two forms, the user might not get that.

> I have not added the submit functionality or validation to the needles form yet. What should happen after submitting the needles form? Should there be a success message? We could use the MUI snackbar if you just want a temporary message. https://material-ui.com/components/snackbars/#snackbar

Snackbar works for me. Working towards better error messages with Yannis, good component for those as well!

frontend/src/components/SepForm.js
LizKalter: Actually, this will throw an invalid prop console warning. The MUI component wants a boolean value.

frontend/src/components/SepForm.js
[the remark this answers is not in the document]
LizKalter: @MikeyManoguerra  I'm getting a 404 from Django on `/api/sites` plural. `/api/site` appears to be returning an empty array. Do I need to do anything to get this populated with data? Thanks.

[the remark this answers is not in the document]
MikeyManoguerra: We just added sites to seed data less than a week ago, so if you have not pulled down lately try that?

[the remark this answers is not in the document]
MikeyManoguerra: @LizKalter

frontend/src/components/SepForm.js
```diff
@@ -106,15 +114,32 @@ const SepForm = observer(() => {
...
+                    name="site_id"
+                    onChange={e => {
+                      handleChange(e)
+                      participantStore.setCurrentSite(e.target.value)
+                    }}
+                    value={values.site_id}
+                  >
+                    {sites.map((site, i) => (
+                      <MenuItem key={i} value={site.id}>
+                        {site.site_name}
```
LizKalter, line 134: I'm currently getting a MUI error because the site_id value is available before the list of sites has been populated. "react_devtools_backend.js:2273 Material-UI: You have provided an out-of-range value `3` for the select (name="site_id") component."

frontend/src/components/SepForm.js
```diff
@@ -46,11 +50,16 @@ const SepForm = observer(() => {
   const classes = useStyles()
   const rootStore = useContext(rootStoreContext)
   const participantStore = rootStore.ParticipantStore
+  const [sites, setSites] = useState([])
```
LizKalter, line 53: I added a local state object for the site list, but I don't think this is the right solution. The Formik form is not technically part of the SepForm component, so the updates to the ParticipantStore object don't appear inside the form. I'm having trouble finding the correct syntax for wrapping the Formik form in observer(). I found this exchange that explains what the issue is, but it does not give me a solution I can directly use. https://github.com/mobxjs/mobx/issues/1505

frontend/src/stores/ParticipantStore.js
```diff
@@ -306,6 +316,48 @@ export class ParticipantStore {
...
+  }) {
+    try {
+      //{"participant":["Incorrect type. Expected pk value, received str."]}
+      const { visitOk, visitData } = yield api.createVisits({
+        program: program,
+        urgency: urgency,
+        participant: participant,
+      })
+      if (!visitOk || !visitData) {
+        throw "a placeholder error string!"
```
LizKalter, line 346: I am getting an error when I call createVisits with these 3 parameters:
{"participant":["Incorrect type. Expected pk value, received str."]}

frontend/src/components/SepForm.js
```diff
@@ -16,19 +27,329 @@ const useStyles = makeStyles({
...
+        validationSchema={SEPNeedleSchema}
+        onSubmit={async (values, { setSubmitting }) => {
+          await participantStore.createSEP({
+            needles_in: values.needles_in,
+            needles_out: values.needles_out,
+            site: participantStore.currentSite,
+            participant: participantId,
+            urgency: 1,
+            program: 10,
+          })
```
LizKalter, line 254: Here is where the values for CreateVisits and CreateSEP are passed. I've left urgency and program hardcoded for now because I am getting an error on the CreateVisits call.
{"participant":["Incorrect type. Expected pk value, received str."]}

frontend/src/components/SepForm.js
MikeyManoguerra: Yes. I think I mentioned this somewhere  back in the issue ( having the site be a mobx piece of state). This should not be an observer connected component at all, actually, if possible. Can you pass down sites array, current site, handle site change method, and form handlers from Sidebar?

frontend/src/components/SepForm.js
```diff
@@ -16,19 +27,329 @@ const useStyles = makeStyles({
...
+          await participantStore.getParticipants({
+            sep_id: values.sep_id,
+            last_name: values.last_name,
+            dob: values.date_of_birth,
+            maiden_name: values.maiden_name,
+          })
+          setSubmitting(false)
+          if (participantStore.participants.length === 1) {
+            setParticipantId(participantStore.participants[0].pp_id)
+            setFieldValue("sep_id", participantStore.participants[0].sep_id)
```
MikeyManoguerra, line 97: The reason you get the participant should id, got str error is you want to `setParticipantId()` to `participantStore.participants[0].id`, not `pp_id`. pp id is preventionpoint's internal id, while just `id` is the database id

frontend/src/components/SepForm.js
MikeyManoguerra: dont worry about this for now.

frontend/src/components/SepForm.js
MikeyManoguerra: i left a comment above addressing this

frontend/src/components/SepForm.js
```diff
@@ -16,19 +27,329 @@ const useStyles = makeStyles({
...
+        validateOnBlur={false}
+        initialValues={{
+          needles_in: "",
+          needles_out: "",
+          visit_date: "",
+        }}
+        validationSchema={SEPNeedleSchema}
+        onSubmit={async (values, { setSubmitting }) => {
+          await participantStore.createSEP({
+            needles_in: values.needles_in,
```
MikeyManoguerra, line 248: this object also requires service. Needle exchange service is 44

frontend/src/stores/ParticipantStore.js
```diff
@@ -306,6 +316,48 @@ export class ParticipantStore {
...
+    program,
+    urgency,
+    participant,
+    needles_in,
+    needles_out,
+    site,
+  }) {
+    try {
+      //{"participant":["Incorrect type. Expected pk value, received str."]}
+      const { visitOk, visitData } = yield api.createVisits({
```
MikeyManoguerra, line 340: This is going to fail, because they are not apiSauce keys
to destructure and rename do this
` const { ok: visitOk, data: visitData } = yield api.createVisits({`

LizKalter: Thanks. After the latest round of changes I'm getting the following error on the createVisits api call:

AttributeError at /api/visits/
'NoneType' object has no attribute 'program'

frontend/src/components/SepForm.js
LizKalter: Thank you. I have updated this.

frontend/src/components/SepForm.js
LizKalter: I have removed the observer wrapper from SepForm in the latest code update, and moved the site related variables to props, but I left the participantStore reference in SepForm as it is used for quite a few other things. It appears to still work okay without the observer wrapper although I don't know if using it this way is considered acceptable.

frontend/src/stores/ParticipantStore.js
MikeyManoguerra: that's a weird error. its actually because it's trying to find service. service has program as a foreign key in the db. null object has no program fk. include service ID in Post

frontend/src/components/SepForm.js
MikeyManoguerra: @LizKalter

frontend/src/stores/ParticipantStore.js
LizKalter: Thanks. That fixed it.

frontend/src/components/SepForm.js
MikeyManoguerra: jk  😅

MikeyManoguerra (changes requested):
You are behind master. other than that it looks good, and we can focus on streamlining UX a bit in followup tickets! Great job

MikeyManoguerra (approved):

PR #419, opened 2020-10-20
Update django admin

JackRyan1989:
Updated SEP_data, Behavioral_Health_Note and HCV_Note models in addition to changing the models listed in admin.py.


core/sep_data/views.py
```diff
@@ -1,7 +1,7 @@
 from core.viewsets import ModelViewSet
-from core.models import Sep_Data
+from core.models import SepData
 from core.sep_data.serializers import Sep_DataSerializer
 
 class Sep_DataViewSet(ModelViewSet):
```
MikeyManoguerra, line 5: Update the view to match the model name

core/sep_data/serializers.py
```diff
@@ -1,10 +1,10 @@
 from rest_framework import serializers
-from core.models import Sep_Data
+from core.models import SepData
 
 
 class Sep_DataSerializer(serializers.ModelSerializer):
```
MikeyManoguerra, line 5: Update the serializer to match the model name

MikeyManoguerra (changes requested):
Looks good! Just a Small naming change.

While looking at the branch, I was reminded you can customize the link in the admin by defining the `__str__` method on the model class. You can see an example on the Visit model.

Lets get this merged in and we can open  a separate  issue for that.

https://docs.djangoproject.com/en/3.1/ref/models/instances/#django.db.models.Model.__str__

JackRyan1989 (commented):
Updated admin.py, sep data model, behavioral health notes model, hcv notes model along with the sep data view and serializer.

JackRyan1989 (commented):
Reviewed and addressed

MikeyManoguerra (approved):
Looks good, Thanks for the final changes

PR #421, opened 2020-10-27
Yup validation for Visit form

JackRyan1989:
Error handling in place, but needs to update state and send to snackbar.

Also, the validation might be a bit excessive when it comes to the programs and services. Since they are drop downs all the when statements are possibly unnecessary.


JackRyan1989:
Error handling with Snackbar in place.

frontend/src/api/visitEndpoints.js
```diff
@@ -2,11 +2,14 @@
 
 export const getVisits = api => async () => await api.get("visits/")
 
-export const updateVisits = api => async (id, body) =>
+export const updateVisits = api => async (id, body) => {
```
MikeyManoguerra, line 5: Was there a specific reason for this?

frontend/src/validation/index.js
```diff
@@ -23,6 +24,147 @@ const searchSchema = Yup.object().shape(
...
+      .integer(),
+  }),
+  program: Yup.number()
+    .required()
+    .positive()
+    .integer()
+    .min(1)
+    .max(10),
+  service: Yup.number()
+    .when("program", {
```
MikeyManoguerra, line 57: This was a lot of work I see, but I'm not sure it's a good idea. I would rather not hardcode the database ids into the front end.  both of these things, and especially services, are subject to addition and removal.  If we really wanted to check this, can we create a helper function that's similar to how the services are filtered for selection? it might be overkill, the backend will deny a mismatch anyway.

frontend/src/components/VisitRouter/index.js
```diff
@@ -38,13 +40,33 @@ const VisitRouter = observer(() => {
...
+          : isValid.map(error =>
+              participantStore.setSnackbarState(
+                `Error location: ${error.name}, Error message: ${error.message}`
+              )
+            )
+      } else {
+        !isValid.length
+          ? participantStore.createVisit()
+          : isValid.map(error =>
+              console.log(
```
MikeyManoguerra, line 60: is this console log supposed to be here?

frontend/src/components/VisitRouter/index.js
```diff
@@ -38,13 +40,33 @@ const VisitRouter = observer(() => {
...
+                participantStore.setSnackbarState(
+                  `Error location: ${error.name}, Error message: ${error.message}`
+                )
+              )
+            )
+      }
+    } catch (err) {
```
MikeyManoguerra, line 67: You can pass this the handleError function. this will I've you a generic "something went wrong" message and then you can put that in the snack bar. 

side note. right now handle error works best for API errors, but we could stand to have a different error handler for non API based errors. that's a TODO though.

frontend/src/components/VisitRouter/index.js
```diff
@@ -38,13 +40,33 @@ const VisitRouter = observer(() => {
...
-    if (existingVisit.id) {
-      participantStore.updateVisit()
-    } else {
-      participantStore.createVisit()
+    try {
+      let isValid = await validateVisitForm(existingVisit)
+      if (existingVisit.id) {
+        !isValid.length
+          ? participantStore.updateVisit()
+          : isValid.map(error =>
```
MikeyManoguerra, line 51: just curious about what happens here with the multiple errors. Does it display them one after the other? how do they come up in sequence?

frontend/src/api/visitEndpoints.js
JackRyan1989: So I can't remember why I made this change, and I think that is a good enough reason for undoing it.

frontend/src/components/VisitRouter/index.js
JackRyan1989: Do you mean pass in the handleError function to the setSnackbarState()?

frontend/src/components/VisitRouter/index.js
JackRyan1989: No sirree!

frontend/src/components/VisitRouter/index.js
JackRyan1989: This is a good catch. I need to fix this. It's not displaying multiple errors.

JackRyan1989: Or well, if you fix the one error and attempt a resubmit it will still display for the next error.

frontend/src/validation/index.js
JackRyan1989: So, yeah I kinda knew this was overkill. I'm going to remove it. I can look into adding a helper function to save us a trip to the backend to validate.

frontend/src/components/VisitRouter/index.js
```diff
@@ -38,13 +41,30 @@ const VisitRouter = observer(() => {
...
+              participantStore.setSnackbarState(
+                `Theres an error in the ${error.name} field.`
+              )
+            )
+      } else {
```
MikeyManoguerra, line 57: change `isValid` to `validationErrors`

i would invert this logic a bit:

"if validationErrors.length 
           return map set snackbar state

if existingvisit.id
  update
else
 create
"

frontend/src/components/VisitRouter/index.js
```diff
@@ -38,13 +41,30 @@ const VisitRouter = observer(() => {
...
+        !isValid.length
+          ? participantStore.createVisit()
+          : isValid.map(error =>
+              participantStore.setSnackbarState(
+                `Theres an error in the ${error.name} field.`
+              )
+            )
+      }
+    } catch (err) {
+      handleError(err)
```
MikeyManoguerra, line 67: handleError returns a string. you should set the snackbar state to this returned string

frontend/src/components/VisitRouter/index.js
MikeyManoguerra: Inverting the logic is not super important, its just a little repetitive. I do want you to change the variable name here at the least though

MikeyManoguerra (changes requested):
Looks good, name change and catch block thing and ill approve. thanks for bearing with me here.

frontend/src/components/VisitRouter/index.js
JackRyan1989: Cool. Yeah I took care of both things. Thanks for catching that.

frontend/src/components/VisitRouter/index.js
JackRyan1989: Addressed in the latest commit.

MikeyManoguerra (approved):

PR #422, opened 2020-11-04
Snackbar

dmjohnspor:


MikeyManoguerra (dismissed):
Looks good! we will eventually want to use the snackbar in all stores, but this will require some constructor finangling so we will do it seperately

MikeyManoguerra:
@dmjohnspor going to have to rebase first

MikeyManoguerra (approved):

PR #430, opened 2020-11-11
Issue 251 all visits related to participant endpoint

alex-cots:
- Added new ParticipantVisitsView, ParticipantVisitsSerializer, ProgramForVisitSerializer
- Added 2 tests

### DIFF
--- a/core/participants/views.py
+++ b/core/participants/views.py
@@ -1,7 +1,12 @@
+from rest_framework.exceptions import NotFound, PermissionDenied
+from rest_framework.response import Response
+from rest_framework.decorators import action
+
 from core.viewsets import ModelViewSet
 from core.models import Participant
 from core.participants.serializers import ParticipantSerializer
-from rest_framework.exceptions import NotFound 
+from core.visits.serializer import ParticipantVisitsSerializer
+
 
 class ParticipantViewSet(ModelViewSet):
     """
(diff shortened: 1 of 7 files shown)


MikeyManoguerra (changes requested):
Sorry for the late response.

My only real request is adding a test for auth (just a no token one works here). currently the front desk doesnt have permission in users_and_groups to get visits, but that will probably change when we finalize the initial permissions in the code.

right now the endpoint throws a 403 forbidden (as expected) when logged in as front desk.  
I guess i have a question for you: since you are not modifying get_queryset, this note on the DRF docs doesnt apply, right? https://www.django-rest-framework.org/api-guide/permissions/#using-with-views-that-do-not-include-a-queryset-attribute

i assume the ListApiView is still using the unmodifyied get_queryset under the hood, even if we havent explicitly defined a queryset on the view

alex-cots:
> I guess i have a question for you: since you are not modifying get_queryset, this note on the DRF docs doesnt apply, right? https://www.django-rest-framework.org/api-guide/permissions/#using-with-views-that-do-not-include-a-queryset-attribute

Ah I think this does apply. I'll test out accessing it without the proper model permissions to confirm first, but I think I need to add the sentinel queryset attribute. I'm assuming this should require Visit model permissions? Or should it be the Participant model since this is really through a participant?

> My only real request is adding a test for auth (just a no token one works here).

Sounds good, I'll add a no token one, and it sounds like one for the proper model permission would be good too.

alex-cots:
Decided to refactor into the ParticipantViewSet using the action decorator. Also added some permissions tests.

It turns out though the DRF's docs are way out of date and as of 3.1.2 the django model perms are decided based off of the get_queryset method if it exists. (https://www.django-rest-framework.org/community/release-notes/#312)

MikeyManoguerra (dismissed):
looks good, and great tests!

MikeyManoguerra:
can you update both serializers with the new field?

alex-cots:
> can you update both serializers with the new field?

I rebased which maybe wasn't the best way to do it but should be added now.

MikeyManoguerra (approved):

PR #434, opened 2020-11-16
Redirect to participant list on SEP search

LizKalter:


MikeyManoguerra (changes requested):
please change 'site' back to 'sites' in the route. Jack just updated the backend, so you may have to rebuild the server?

also theres some wierd overwriting going on with the different branches in participantStore. can you make sure your file matches what is currently on master after rebase?

otherwise looks good

MikeyManoguerra (approved):

PR #436, opened 2020-11-17
Participation form validation 426

JackRyan1989:
Schema validation ready for review.


MikeyManoguerra:
hey hang tight for a minute, i want to merge these in one at time bc there was some overwriting going on

frontend/src/validation/index.js
```diff
@@ -62,9 +103,42 @@ const visitSchema = Yup.object().shape({
...
+    then: Yup.string().notRequired(),
+    otherwise: Yup.number().required(),
+  }),
+})
+
+const validateForm = (data, inputSchema) => {
   let errors = []
-  return visitSchema
+  let schema
+  if (inputSchema === "visitSchema") {
```
MikeyManoguerra, line 136: i like this.

Can you make constants for the schema names?  `const VISIT_SCHEMA = 'VISIT_SCHEMA'` and then import it from here to the visitRouter/index.

frontend/src/validation/index.js
```diff
@@ -62,9 +103,42 @@ const visitSchema = Yup.object().shape({
...
-const validateVisitForm = data => {
+// eslint-disable-next-line no-unused-vars
+const participantSchema = Yup.object().shape({
+  first_name: Yup.string()
+    .required()
+    .max(50),
+  last_name: Yup.string()
+    .required()
+    .max(100),
+  date_of_birth: Yup.string().required(),
```
MikeyManoguerra, line 114: Id like to make these dates valid for a reasonable period, between 1900 and today. i messed around with it and we have to turn the date string back into a date object, or extend the schema and handle it ourselves. 

idea is to add a new case to handleParticipant update fn, that just dumps the date string `yyyy-MM-dd` into new Date(),  then update this to yup.date() with min and max.

i tested it and it submitted successfully to the back end, though it seems a little silly? /shrug

MikeyManoguerra (changes requested):
looks good. the one thing for sure i want is the constants. we can talk about the date validation if you like. 

Just a heads up, there is another participant form in existing participant view. if you can add that here without to much trouble, that woudl be cool, otherwise we do it later.

also, rebase.

JackRyan1989:
@MikeyManoguerra So I'm running into a slight issue when validating the existing participant form. When creating a participant the 'insurer' field is an empty string when 'is_insured' is set to false, otherwise it is a number. However, on for an existing participant you can set the 'is_insured' field to false, but the 'insurer' field can remain as a number. I'm not sure if we should pursue setting the 'insurer' field to an empty string when 'is_insured' is false or if I should try some clever work around for validation, thoughts?

JackRyan1989:
@MikeyManoguerra So actually there's a tricky thing in general which is that when you are creating a new participant or changing a field on an existing participant the value goes through as a string, but when that information gets pulled in from the backend it's often times a number. A good example of this is the SEP ID field, when you create it it's a string. When it comes to the frontend it's there as a number, but then once you update it it's a string again. This is making validation of the existing participant form quite difficult. I think we need to decide on keeping values either as strings on the front end, or not worry about validation for some of the fields that change type.

MikeyManoguerra:
if you can set the insurer field to empty sting that's the way to go.  It might break the back end update logic. i don't know how the backend handles an empty field for insurer when it was set before. I think thats the implied logic by switching from t to f, they no longer have insurance.  (there won't be a record over time of their changing status, but that's ok for now).

Yea and the number to string issue is going to happen since we are storing the values as inputs. im fine with switching the value to a string before submitting. as long as the update goes through 200

JackRyan1989 (commented):
Changes made

JackRyan1989:
@MikeyManoguerra I changed the sep_data model to be a charfield. Let me know if that is no bueno.

MikeyManoguerra:
So my knee jerk reaction is to say id rather change it on the front end.  but the reason is to maintain numerical integrity from a data perspective.

if stay like this you should do something similar to the clean method i wrote here : https://github.com/CodeForPhilly/prevention-point/blob/master/core/models/visit.py, but compare the incoming value to a regex.

if on the front end, when an existing participant is passed to the `setParticipant` method,  you could 
do something like 
```
@action setParticipant(data){
const {sep} =data
this .participant = {...data, sep: sep ? sep.toString(): ''}
}
```
with the ternary making it extra cautious haha

I feel like the whole thing is going to get an audit from an UX and data perspective, and this might not matter/be wrong. the ssn on this table is char.

MikeyManoguerra:
@JackRyan1989 https://docs.djangoproject.com/en/3.1/ref/models/instances/#validating-objects

JackRyan1989:
@MikeyManoguerra Great suggestion with the ternary. And yeah, I felt not great (smelly) about changing a model to match our validation on the frontend. 
Tested both the update and new participant validation and all worked well. Let me know if you see anything else that should be changed.
Happy Thanksgiving btw!

MikeyManoguerra:
Happy thanksgiving to you too! Looks good, The only issue i saw was that when the is_insured is false, theres still an insurer FK. but this is a seed / model validation issue, and should be updated there. looks good!

PR #439, opened 2020-11-19
SEP form date default to current day

LizKalter:
Please note that the date value is not currently passed when calling createSEP().


MikeyManoguerra (changes requested):
Can you pass the value to create SEP? send the same value to the visit and sep POST. if there's a formatting error, there's a date formatting util `// utils` in participantStore.

also, you can use the snackbar in the endpoint instead of the placeholder error.

MikeyManoguerra (approved):

PR #440, opened 2020-11-26
restructure store provider context, update tests to reflect change

MikeyManoguerra:
@dmjohnspor  this is a branch to get us started on testing. The change i made was in how the application inititalizes the Store, which allows us to manually set a mock store before passing it to the test. check this out and look at it, and then we can talk about writing some more tests. are you free this weekend ( the next four days?)


MikeyManoguerra:
merged under another name

PR #452, opened 2020-11-27
Separate participants store 425

JackRyan1989:
@MikeyManoguerra So I created a new store titled "NotificationStore". Currently it just handles the snackbar state. Is there anything else I should move over to this new store from the participant store? Maybe I should make the error message const an observable? 

Also, I can title it something else if need be. Was thinking UtilityStore, ErrorStore or GeneralStore (har har) as options but none felt totally accurate.


MikeyManoguerra:
general store is funny. I like utility store, use that.

you can move drawer logic and and search observable and their logic over.

error const doesn't need to be bc it is just a string id basically.


one other thing, is it possible to reference the participantstore (or any non root store) in the constructor? if you can't find an easy answer it's no big deal but it would be better not to have to drill down so much

JackRyan1989:
@MikeyManoguerra I renamed the new store to UtilityStore as recommended. I also moved over the search sidebar stuff. 

Weirdly, I couldn't get the UtilityStore to work when put into the constructor. The error I got was that that rootStore was not defined. Which is throwing me for a loop given this stack overflow exchange: https://stackoverflow.com/questions/44928645/how-can-i-access-mobx-store-in-another-mobx-store#44928890

Not sure if it has to do with how we're passing the context in or not.

frontend/src/components/VisitRouter/index.js
```diff
@@ -48,7 +49,7 @@ const VisitRouter = observer(() => {
       let validationErrors = await validateForm(existingVisit, VISIT_SCHEMA)
       if (validationErrors.length) {
         return validationErrors.map(error =>
-          participantStore.setSnackbarState(
+          notificationStore.setSnackbarState(
```
MikeyManoguerra, line 52: missed an update here

frontend/src/components/VisitRouter/index.js
```diff
@@ -59,7 +60,7 @@ const VisitRouter = observer(() => {
         participantStore.createVisit()
       }
     } catch (err) {
-      participantStore.setSnackbarState(handleError(err))
+      notificationStore.setSnackbarState(handleError(err))
```
MikeyManoguerra, line 63: and here

MikeyManoguerra (changes requested):
Looks good except for the old name in the visit router.

frontend/src/components/VisitRouter/index.js
JackRyan1989: Updated, thanks for catching!

frontend/src/components/VisitRouter/index.js
JackRyan1989: Updated, thanks for catching!

MikeyManoguerra (approved):

PR #453, opened 2020-11-29
Issue 431 and 441 - snackbar updates

LizKalter:
This PR updates the Snackbar component to use MUI Alert and updates the error handling to include the severity level for the Alert component. The SEP form has also been updated to display a success message using the Snackbar and clear the form.

### DIFF
--- a/frontend/src/components/SepForm.js
+++ b/frontend/src/components/SepForm.js
@@ -258,8 +258,7 @@ const SepForm = ({ sites, currentSite, setCurrentSite }) => {
           setSubmitting(false)
           if (success) {
             handleClear()
-            utilityStore.setSnackbarState({
-              message: "SEP data submitted",
+            utilityStore.setSnackbarState("SEP data submitted", {
               severity: SNACKBAR_SEVERITY.SUCCESS,
             })
           }
--- a/frontend/src/components/SepForm.js
+++ b/frontend/src/components/SepForm.js
@@ -15,6 +15,7 @@ import PrevPointHeading from "./Typography/PrevPointHeading"
 import PrevPointCopy from "./Typography/PrevPointCopy"
 import { Formik, Form } from "formik"
 import { SEPSearchSchema, SEPNeedleSchema } from "../validation"
+import { SNACKBAR_SEVERITY } from "../constants"
 
 const useStyles = makeStyles(theme => ({
   root: {
@@ -243,7 +244,7 @@ const SepForm = ({ sites, currentSite, setCurrentSite }) => {
         }}
         validationSchema={SEPNeedleSchema}
         onSubmit={async (values, { setSubmitting }) => {
-          await participantStore.createSEP({
+          const success = await participantStore.createSEP({
             needles_in: values.needles_in,
             needles_out: values.needles_out,
             visit_date: values.visit_date,
@@ -254,6 +255,13 @@ const SepForm = ({ sites, currentSite, setCurrentSite }) => {
             service: 44,
           })
           setSubmitting(false)
+          if (success) {
+            handleClear()
+            participantStore.setSnackbarState({
+              message: "SEP data submitted",
+              severity: SNACKBAR_SEVERITY.SUCCESS,
+            })
+          }
         }}
       >
         {({ errors, touched, values, handleChange, isSubmitting }) => (
--- a/frontend/src/components/SepForm.js
+++ b/frontend/src/components/SepForm.js
@@ -52,6 +52,7 @@ const SepForm = ({ sites, currentSite, setCurrentSite }) => {
   const classes = useStyles()
   const rootStore = useContext(rootStoreContext)
   const participantStore = rootStore.ParticipantStore
+  const utilityStore = rootStore.UtilityStore
   const [participantId, setParticipantId] = useState()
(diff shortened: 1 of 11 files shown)


frontend/src/stores/ParticipantStore.js
```diff
@@ -161,10 +165,17 @@ export class ParticipantStore {
...
-  @action setSnackbarState = (message, options = { open: true }) => {
-    // we setup otions like this to maybe add more later
-    this.snackbarState.message = message
-    this.snackbarState.open = options.open
+  @action setSnackbarState = ({
+    message = "",
+    severity = SNACKBAR_SEVERITY.INFO,
+    open = true,
+  }) => {
+    // prevent visual change to severity or message when closing the snackbar
```
MikeyManoguerra, line 173: was the message /severity changing before it disappeared?

frontend/src/stores/ParticipantStore.js
```diff
@@ -161,10 +165,17 @@ export class ParticipantStore {
   @action setSidebarView = sidebarView => {
     this.sidebarView = sidebarView
   }
-  @action setSnackbarState = (message, options = { open: true }) => {
-    // we setup otions like this to maybe add more later
-    this.snackbarState.message = message
-    this.snackbarState.open = options.open
+  @action setSnackbarState = ({
```
MikeyManoguerra, line 168: I like this approach better; (actually it was implemented like this from where i took it but in typescript). I designed it the other way  (with message, then options)  so it was a little easier to use and harder to send an empty message.  Since we are not using typescript, was there a reason you had to switch it?

I'm ok with leaving it because ultimately we wont be doing to much more. It would be cool to have generic "success", "something went wrong", etc messages as defaults instead of an empty string, but then we'd have to conditionally pick them based on the enum.

yea but curious why you changed it

frontend/src/error/index.js
```diff
@@ -33,4 +40,20 @@ function handleError(statusCode) {
...
+      severity: SNACKBAR_SEVERITY.ERROR,
+    }
+  }
+  return {
+    message: standardizedError.description,
+    severity: standardizedError.severity,
+  }
+}
+
 export default handleError
```
MikeyManoguerra, line 59: oo I dont remember this sneaking in, can you just export the handle error fn, we dont need a default export.

MikeyManoguerra:
Looks good overall, just a couple of questions.
Also hold tight a sec, Im going to have you wait to merge this until jack merges his store PR. 
He moved the snackbar method to a new UtilityStore. I think itll be easier for you to reconcile with his changes then he move yours.

frontend/src/stores/ParticipantStore.js
LizKalter: Yes. There is an animation when it disappears, but the severity and message change happens instantly.

frontend/src/stores/ParticipantStore.js
LizKalter: The deconstruction of the options object properties with default values doesn't work when the options object is a second parameter. All of the parameters have to get passed as a single object for this to work.

frontend/src/error/index.js
LizKalter: Done.

MikeyManoguerra:
@LizKalter you are set to rebase this. let me know if there are any issues with the methods moving over

frontend/src/stores/ParticipantStore.js
MikeyManoguerra:
```
  @action setSnackbarState = (message, options = {}) => {
    const severity = options.severity ? options.severity : SNACKBAR_SEVERITY.ERROR
    const open = options.open ? options.open : true
    // prevent visual change to severity or message when closing the snackbar
    if (open) {
      this.snackbarState.severity = severity
      this.snackbarState.message = message
    }
    this.snackbarState.open = open
  }
```
heres what i was talking about in the meeting
id rather have it like this  just to keep it easier to use, since there might have already been an mistake/error somewhere else lol

MikeyManoguerra (commented):
heres the link.
https://formik.org/docs/api/formik#resetform-nextstate-partialformikstatevalues--void

you can have handle clear have a nextState parameter with a default of null
then 

do something like

```
if(!nextState) {
return clearForm()
}
clearForm(nextState)
```

LizKalter:
@MikeyManoguerra  Sorry, I am not sure what change you are looking for with regard to the nextState parameter. What is the behavior you are looking for when the form is cleared? (I noticed the needles fields weren't being cleared and added those.)

MikeyManoguerra:
@LizKalter The Site field should not clear

frontend/src/components/SepForm.js
```diff
@@ -65,7 +65,12 @@ const SepForm = ({ sites, currentSite, setCurrentSite }) => {
   const handleClear = () => {
     setParticipantId(null)
     if (SEPParticipantFormRef.current) {
-      SEPParticipantFormRef.current.resetForm()
+      SEPParticipantFormRef.current.resetForm({
+        values: {
+          ...SEPParticipantFormRef.current.initialValues,
```
MikeyManoguerra, line 70: nice!

MikeyManoguerra (approved):

PR #454, opened 2020-11-29
String methods 446

JackRyan1989:
Added __str__ method to 4 models that currently appear in the admin site.


MikeyManoguerra:
looks good. if you dont know about f strings, they are way easier, (like ${}) but its all the same to me!

MikeyManoguerra (approved):

PR #456, opened 2020-12-04
Insurer issue 448

JackRyan1989:
@MikeyManoguerra So this is not working currently, but I think it's close!

I'm running into an issue in the participant model. 

In the cleaning function (def clean(self), line 49 on) I'm having trouble with two things.

The seed.py script does not run if `self.insurer.name != ""` is not in the participant model. However, validation does not work with this setting. 
Here is the error: 
`File "/Users/jr3567/Desktop/prevention-point/core/models/participant.py", line 54, in clean
    if self.is_insured == (False) and (self.insurer.name != ""):
AttributeError: 'NoneType' object has no attribute 'name'`

All I'm trying to do in the clean function is have access to the name attribute of the Insurer model. I'm not sure that that's possible? I've tried looking through the docs a little bit but haven't found anything helpful.

On the front end, an additional blank space appears in the insurer drop down. Not sure that is an issue, but wanted you to be aware.


MikeyManoguerra:
Ill have to study it a little closer for a full answer, but it seems weird to give the `insurer.name` a `blank ="true"`.  the participant model should be the only one affected here.

JackRyan1989:
`but it seems weird to give the insurer.name a blank ="true"` 
So I did that. That was to allow the name field to have an empty string when the user sets the "is_insured" field to false. I think the Insurer model in the Participant model is only referring to the FK, so wouldn't allowing that to be blank wouldn't affect the underlying Insurer model, right?

MikeyManoguerra:
Yes, that is right. I think it might be more clear if you set the FK to accept null, not blank? If a participant doesnt have an insurer, the field should just be empty, not reference a "'no insurer' insurer".

maybe the easiest way to handle this is to only allow an insurer FK to saved if 'is_insured' is true. if 'is_insured' false, the clean method can set it to blank/null before saving

JackRyan1989:
@MikeyManoguerra so I changed the participant model to only work on the FK. This removed any need to update the seed.py script. Weirdly, the backend tests would fail because the maiden_name parameter wouldn't accept blanks? So I set blank to true for that one along with the insurer field because I got "field cannot be blank" errors there too. I don't see this as an issue so long as we are careful of what we send from the front end. Plus, the insurer field needs to accept blanks because the frontend sends back an empty string rather than null, which is what it is coerced to by the clean function in the model. Sorry if this is garbled. I tired.

MikeyManoguerra (approved):

Issue #443, opened 2020-11-27
Improve Visit Router


MikeyManoguerra:
new path `/existingparticipant/<participant-id>/`
on page load, if participant id  in path does not match participantStore.participant.id, request path participant from api. 

change redirect to 404 page if participant not found.

MikeyManoguerra:
If anyone else wants to take this let me know, but I think I can knock it out pretty quick

PR #459, opened 2020-12-05
enhance front end visit router to check params for particpant id

MikeyManoguerra:
This makes the  existing participant path a little more resilient, and can handle a a refresh without redirecting away from the path.

This adds the participant id, and visit id, into the front end paths.

I also changed the path from `/existingParticipant/` to `/participants/:participantId/` to be more RESTful

#443 

The visitTable changes are just Lorem Ipsum filler, the big changes are the router paths

### DIFF
--- a/frontend/src/views/ExistingParticipantView.js
+++ b/frontend/src/views/ExistingParticipantView.js
@@ -53,7 +53,7 @@ const ExistingParticipantView = observer(() => {
     participantStore.getInsurers()
     participantStore.getPrograms()
     async function matchParticpantIdToParams() {
-      if (!participantStore.participant.id !== participantId) {
+      if (participantStore.participant.id !== parseInt(participantId, 10)) {
         // get participant data via param id if not in state or mismatch between state and route
         const ok = await participantStore.getParticipant(participantId)
         if (!ok) {
--- a/frontend/src/views/ExistingParticipantView.js
+++ b/frontend/src/views/ExistingParticipantView.js
@@ -1,7 +1,7 @@
 import React, { useContext, useEffect } from "react"
 import { toJS } from "mobx"
 import { observer } from "mobx-react-lite"
-import { useHistory, Link } from "react-router-dom"
+import { useHistory, Link, useParams } from "react-router-dom"
 import { makeStyles } from "@material-ui/styles"
 import Grid from "@material-ui/core/Grid"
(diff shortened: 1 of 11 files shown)


frontend/src/views/ExistingParticipantView.js
```diff
@@ -41,25 +41,30 @@ const ExistingParticipantView = observer(() => {
...
+        const ok = await participantStore.getParticipant(participantId)
+        if (!ok) {
+          participantStore.setDefaultParticipant()
+          participantStore.setDefaultVisit()
+          history.push("/404/")
+        }
+      }
     }
+
+    matchParticpantIdToParams()
```
JackRyan1989, line 67: Where is this function defined?

MikeyManoguerra: within the useEffect block itself. the callback given to useeffect cant be async

frontend/test/PrevPointTableBody.test.js
```diff
@@ -53,7 +56,10 @@ describe("<PrevPointTableBody />", () => {
     const { getByText } = render(
       <BrowserRouter>
         <PrevPointTableBody participants={[]} />
-      </BrowserRouter>
+      </BrowserRouter>,
+      {
+        container: document.body.appendChild(table),
```
JackRyan1989, line 61: Is this for the useRef()?

frontend/test/PrevPointTableHead.test.js
```diff
@@ -9,11 +9,13 @@ const tableContainer = {
...
 describe("<PrevPointTableHead />", () => {
   it("should render a PrevPointTableHead component", () => {
-    render(<PrevPointTableHead />)
+    render(<PrevPointTableHead />, tableContainer)
   })
 
   it("should render a PrevPointTableHead and have a TableHead element", () => {
-    const { getByLabelText } = render(<PrevPointTableHead />, tableContainer)
+    const { getByLabelText } = render(<PrevPointTableHead />, {
+      container: document.body.appendChild(table),
```
JackRyan1989, line 17: Why is this table being appended?

JackRyan1989 (commented):
Some small questions. I think this looks good. I'm not sure what other tests you would want to write or whether you would need to at the moment.

JackRyan1989 (approved):

PR #460, opened 2020-12-06
Has queue flag 444

JackRyan1989:
Added "has_queue" flag to programs model.
Updated programs fixture.
Made new migration.
Updated tests.


core/tests/programs.py
```diff
@@ -92,3 +92,16 @@ def test_is_frozen_is_updated(self):
         ).values()[0]
 
         self.assertTrue(updated_program['is_frozen'])
+
+    def filter_by_queue(self):
```
MikeyManoguerra, line 96: did this test run? i think tests have to start with "test"?

MikeyManoguerra: `pipenv run test core.tests.ProgramsTestCase`

MikeyManoguerra (changes requested):
needs a rebase, and also the namespace question

core/tests/programs.py
JackRyan1989: Now that you mention it, I don't think it did. There was the same number of tests after I put this in. I thought that was strange, shrugged my shoulders and committed because it passed. Ok cool, I'll fix and resubmit.

MikeyManoguerra (approved):

MikeyManoguerra (approved):

PR #461, opened 2020-12-07
Populate SEP form from table

LizKalter:


MikeyManoguerra (approved):
Looks good. 
The only thing i want to add is  a useeeffect that clears the sep form on ummount
```
  useEffect(() => {
    return handleClear
  }, [])
```

so if you switch between search tabs or navigate to existing participant and back , the form is cleared. put this in and ill approve

Seperately, i think the SEPform component is now big enough to split into multiple files, but not exactly sure how yet. just throwing that out there

MikeyManoguerra (approved):

PR #468, opened 2020-12-16
First pass at filtering.

JackRyan1989:
Needed to update seed script so some programs come through. Also set has queue to True initially, I think we want folks to manually set the program to not have a queue that intuitively makes sense to me.


core/management/commands/seed.py
```diff
@@ -239,6 +239,7 @@ def create_programs(output=True):
     for program, services in DEFAULT_PROGRAMS.items():
         p = Program()
         p.name = program
+        p.has_queue = random_bool()
```
MikeyManoguerra, line 242: Even though we aren't using the queue right now, which programs have queues should not be random. 
We definitely do not want SEP to have a queue at the moment. 

Create an list of strings const called HAS_QUEUE, with a few program names "STEP" , "LEGAL" and a couple more. then if the program is in the list while seeding, make the has queue flag true

frontend/src/views/AllQueues.js
```diff
@@ -52,9 +52,17 @@ const AllQueues = observer(() => {
     api
       .getPrograms()
       .then(({ data }) => {
```
MikeyManoguerra, line 54: we should run the filter before we make the requests to the queue endpoint. That endpoint is relatively expensive, and nothing will come back anyway.

try using Array.filter method, itll be a little more clear what is going on

MikeyManoguerra (changes requested):
just a couple small things

frontend/src/views/AllQueues.js
JackRyan1989: So do you think where I'm implementing the filter is ok @MikeyManoguerra ?

MikeyManoguerra: Yes, If its something that a more common operation we could seperate it out into a fn, but this is the only place we are doing it so far

MikeyManoguerra (approved):

PR #469, opened 2020-12-20
Add loading state spinner

LizKalter:
I updated all of the references to the QueueStore to use RootStoreContext because otherwise the reference to UtilityStore within QueueStore did not exist. If this change causes problems we will need some other way to reference UtilityStore within QueueStore.

I am referring to:
`const rootStore = useContext(RootStoreContext)`
`const queueStore = rootStore.QueueStore`
vs
`const queueStore = useContext(QueueStoreContext)`
so that the following works within QueueStore:
`constructor(rootStore) {
    this.rootStore = rootStore
  }`


LizKalter:
@MikeyManoguerra AllQueues.test.js is getting the same error where the RootStore is undefined within QueueStore that I encountered above, but I am not sure what the fix is here.

MikeyManoguerra:
It looks like the tests failed on here due to a linter issue, I pulled the branch down and they passed locally.  I think I have had this issue before, (with the way the allqueues component uses the store) but I cant reproduce at the moment???

LizKalter:
@MikeyManoguerra I didn't have the latest from master merged back into my local branch, so my linter and test results were a red herring. I think all is well now.

frontend/src/layouts/MainLayout.js
```diff
@@ -19,6 +19,7 @@ const MainLayout = observer(props => {
     drawer: {
       width: drawerWidth,
       flexShrink: 0,
+      zIndex: 1,
```
MikeyManoguerra, line 22: do we need to set this? it locks the stacking context, and we should only do so if necessary.

MikeyManoguerra (commented):
Id prefer not to set zindex on the layout, bc then basically anything that is in that component (basically everything) is locked at a level below the nav-bar. If its needed its fine but if not Id like to remove it as to not cause consternation for someone in the future. ( can you tell ive ran into this issue before on another project?)

MikeyManoguerra (dismissed):

frontend/src/layouts/MainLayout.js
LizKalter: No, but we still need to apply z-index somewhere in order to get the spinner overlay to appear above the header and drawer. MUI has z-index values applied to the header and drawer by default, so I was overriding them. Instead I've moved the z-index property to the MainLayout div. This might be a better choice because it just enforces a stacking order between the main div and the snackbar/spinner which should always be over top of it anyway.

MikeyManoguerra (approved):

PR #470, opened 2020-12-21
use correct id for populatinon

MikeyManoguerra:
@LizKalter found this, using the wrong id when populating from table. I missed it in review. The spreading out of "Other values" might be overkill, but i did not want to screw up what values are being passed to formik out of caution

### DIFF
--- a/frontend/src/components/SepForm.js
+++ b/frontend/src/components/SepForm.js
@@ -64,6 +64,8 @@ const SepForm = ({
   const SEPNeedleFormRef = useRef()
   const history = useHistory()
 
+  const { participant_id, ...OtherSEPValues } = SEPFormValues
+
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


frontend/src/components/SepForm.js
```diff
@@ -77,16 +79,18 @@ const SepForm = ({
...
+          ...OtherSEPValues,
           site_id: SEPParticipantFormRef.current.values.site_id,
         },
       })
     }
     if (SEPNeedleFormRef.current) {
       SEPNeedleFormRef.current.resetForm()
     }
-    setParticipantId(SEPFormValues.sep_id)
-  }, [SEPFormValues])
```
LizKalter, line 89: Oops! Sorry about that!

LizKalter (approved):

PR #471, opened 2020-12-28
Implemented snackbar on auth and queue stores.

JackRyan1989:
Couple o' q's:

1.  The handleError fn is not generating a message or severity properties with the errors I am passing in via auth and queue stores. As a result, I'm sending just the variable "snackbarError", which is a string, to the setSnackBar state action. For severity, I'm setting that to a string, "error", manually. Is handleError going to be worked on so it will always generate  message and severity properties? If so, I can just put those in place in preparation for those changes.
2. I noticed that Notification store has a setSnackBar state action as well that doesn't have severity integrated. Should that action be removed?

Thanks!


MikeyManoguerra:
The issue here is you are using `handleError` and not `handleSnackbarError` from the error handlers module.

This is definitely confusing setup however.

Really, `handleSnackbarError`  should be `handleApiError`, and `handleError` should be removed. This error mapping normalizer function only works when its passed a status code, so client/browser only errors using the snackbar should not use this this function.

 the can be handled on a case by case basis, in lieu of writing a client error handler mapper

frontend/src/stores/ParticipantStore.js
```diff
@@ -383,7 +383,7 @@ export class ParticipantStore {
       }
       this.setRouteToQueue(true)
     } catch (error) {
-      const snackbarError = handleSnackbarError(error.message)
```
MikeyManoguerra, line 386: These should stay as they are. Error is an object, and whatever you pass to the constructor is available as value of a `message` key. `new Error(<accessable at key 'message'>)` its stored as a string, which is why in the error/index.js file we are parsing the integer from the parameter

MikeyManoguerra (changes requested):

MikeyManoguerra (approved):

PR #483, opened 2021-01-14
Insurer field bug

### LINKED ISSUE #480: insurer field errors
Yup does not want you to mix types for a single field,  which i  s the reason for the generic ” something went wrong” error getting kicked to the catch block.

 ```
is_insured: Yup.boolean().required(),
  insurer: Yup.string().when("is_insured", {
    is: false,
    then: Yup.string().notRequired(),
    otherwise: Yup.number().required(),
  }
```

The schema isnt returning an error the schema is the error, so to speak. this is intermittent if i remember  correctly, based upon what the value passed is 

___
The select dropdown is maintaining the value as a number for insurers, which is weird behavior because a value should always be a string i thought, even when all the characters are integers 

anyway… the backend can handle a type string number as the value for insurer_id, so i think we should just coerce all the values to strings on the front end, and let the backend deal with it.

- on submit, make sure the value of `participantStore.visit.insurer` is coerced to a string
- try to use Yups `.nullable()` (i havent tested it) if not, check for `insurers: null` in the object  on submit and instead send and swap it for an  empty string before sending it to yup
- if `is _insured` is set to false, make sure no id is retained in the `insurers` field

#### compare by reference, not by value here:

`if (inputSchema === "VISIT_SCHEMA") {`

should be

`if (inputSchema === VISIT_SCHEMA) {`

in `validation/index`

JackRyan1989:
Insurer field only goes through in Yup if set to notRequired. Validates properly when an insurer is present but does not validate if 'is_insured' is false. Because we're setting the value for insurer in the participantstore to empty string and then converting to null on the backend I think this is ok. But let me know if this gives you the heebie jeebies.

Closes #480

### DIFF
--- a/frontend/src/stores/ParticipantStore.js
+++ b/frontend/src/stores/ParticipantStore.js
@@ -105,11 +105,14 @@ export class ParticipantStore {
       case "is_insured":
         this.participant.is_insured = value === "true"
         if (!this.participant.is_insured) {
-          this.participant.insurer = ""
+          this.participant.insurer = " "
         }
         break
       case "pp_id":
         this.setPPId(value)
+        break
+      case "insurer":
+        this.setInsurer(value)
         break
       default:
         this.participant[name] = value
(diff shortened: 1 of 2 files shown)


MikeyManoguerra (approved):
yep i think this is the right way to handle it.

PR #495, opened 2021-01-25
Slugs for services 447

MikeyManoguerra:
This adds a slug field for the service.  
 the django admin allows the user to create a slug field, but then it is un-editable.
its also un-editable in the serializer, which means its set for life once created.

The result is we can use this as a unique identifier for services. we can also use this to create dynamic route requests in the future, which will help for an editable form


MikeyManoguerra:
i am making this as a draft bc i want to rebase once jack gets his most recent PR merged in. then i will wipe the migrations, since i edited them directly to get this working in development.

JackRyan1989:
@MikeyManoguerra I have merged in my latest PR. I'll wait for you to wipe the migrations and then I'll take a look at this review.

MikeyManoguerra:
@JackRyan1989 ok its ready

core/models/program.py
```diff
@@ -1,5 +1,5 @@
 from django.db import models
-
+import re
```
JackRyan1989, line 2: Why import re?

core/programs/views.py
```diff
@@ -1,6 +1,6 @@
 from core.viewsets import ModelViewSet
 from rest_framework.response import Response
-from core.models import Program, Service
+from core.models import Program, Program
```
JackRyan1989, line 3: Why import Program twice?

JackRyan1989 (approved):
Two questions, neither are critical nor should they hold up this PR.

core/management/commands/seed.py
```diff
@@ -253,13 +254,15 @@ def create_programs(output=True):
...
+            s.slug = f"{p.name}-{s.name}"
+
             s.available = random_bool()
             s.program = p
             s.full_clean()
             s.save()
 
             if output:
-                print("Created {}: '{}'".format(p.name, s.name))
+                print("Created {}: '{}'".format(p.slug, s.slug))
```
alex-cots, line 265: I might be missing something but I don't think `p` has a slug attribute

alex-cots (changes requested):
Backend looks good. I think just the one line in seed is a small mistake, plus Jack's comments

core/management/commands/seed.py
MikeyManoguerra: ah dang you're right, i don't know how that slipped back in.

core/programs/views.py
MikeyManoguerra: weird no idea

core/models/program.py
MikeyManoguerra: i was going to add a progam slug then decided against it. will remove

alex-cots (approved):

JackRyan1989 (approved):

PR #501, opened 2021-02-02
Alphanumeric sep id 497

### LINKED ISSUE #497: sep_id on participant needs to be  alphanumeric
right now its just an integer field. change to varchar
PS: i wrote an extra amount of tests for this. Knowing Prevention Point's propensity for change, we can probably remove the character count restrictions.

JackRyan1989:
Enforces lowercaseness for SEP_ID search and creation. UI Test is present, seed script has been modified, all unnecessary migrations have been deleted, changes to front end mostly revolve around enforcing lowercase sep ids, participant model has been modified to make SEP ID a varchar and be indexed. Maiden Name and Date of Birth are also indexed in the DB since they are searchable on the front end. 

Closes #497


MikeyManoguerra (approved):

MikeyManoguerra:
i know it was there but i'm anticipating we have to remove the unique flag 🤷 . just leave it for now

PR #503, opened 2021-02-10
feat(ci): kubernetes deployment

themightychris:
Don't have this working yet, running into weird auth problems with my local kind cluster pulling from ghcr.io even though I had a `regcred` secret in place that looked good

This _might_ be enough for a working deploy after that, but I have no idea if my use of `hostAliases` will make the `app` and `db` hostnames hardcoded into the frontend nginx config and backend will work. Ultimately we want the backend connecting to a database provided through env, and to probably let k8s' ingress handle mounting `/api`

TODO:
- [ ] Remove line from Dockerfile
- [ ] Mount `/` with docker-compose
- [ ] Create ingress path for container on `/api`
- [ ] *Django App*: Receive DB Hostname and Creds through ENV

### DIFF
--- a/helm-chart/values.yaml
+++ b/helm-chart/values.yaml
@@ -1,0 +1,63 @@
+# Default values for prevention-point.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+backend:
+  image:
+    repository: ghcr.io/codeforphilly/prevention-point/backend
+    pullPolicy: Always
+    # Overrides the image tag whose default is the chart appVersion.
+    tag: "latest"
+
+frontend:
+  image:
+    repository: ghcr.io/codeforphilly/prevention-point/frontend
+    pullPolicy: Always
+    # Overrides the image tag whose default is the chart appVersion.
+    tag: "latest"
+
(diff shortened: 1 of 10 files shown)


themightychris:
ah ha, I've never used ghcr.io before, you have to explicitly switch a package to public after you publish it

themightychris:
we should probably remove the line i added to dockerfile to copy nginx config in. If the container build without that provides for serving the static content out of /, then that can be something we just mount with docker-compose and the k8s ingress can handle plugging the api into the `/api` path on top of it

helm-chart/values.yaml
```diff
@@ -0,0 +1,63 @@
...
+podAnnotations: {}
+
+podSecurityContext: {}
+  # fsGroup: 2000
+
+service:
+  type: ClusterIP
+  port: 80
+
+ingress:
```
MooseQuest, line 46: Do we need to clean this up, or is just fine lingering here disabled?

MooseQuest: Will enable and add correct values via different branch

MooseQuest (approved):
Approving and Merging

